PR #17, opened 2019-06-21
U/wmwv/download data suggestions

wmwv:
@djperrefort Here are just a few suggestions for how I would have done some small things differently in the code in data_download.py

### DIFF
--- a/broker/ztf_archive/_download_data.py
+++ b/broker/ztf_archive/_download_data.py
@@ -61,8 +61,8 @@ def get_local_release_list():
     if not os.path.exists(ALERT_LOG):
         return []
 
-    with open(ALERT_LOG, 'r') as ofile:
-        return [line.strip() for line in ofile]
+    with open(ALERT_LOG, 'r') as infile:
+        return [line.strip() for line in infile]
 
 
 def get_local_alert_list():
--- a/broker/ztf_archive/_download_data.py
+++ b/broker/ztf_archive/_download_data.py
@@ -17,7 +17,7 @@ from tqdm import tqdm
 FILE_DIR = Path(__file__).resolve().parent
 DATA_DIR = os.path.join(FILE_DIR, 'data')
 ALERT_LOG = os.path.join(DATA_DIR, 'alert_log.txt')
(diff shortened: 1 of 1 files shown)


djperrefort:
This all looks good to me, but I’m interested to know why you prefer `os.path` over `Pathlib`?

wmwv:
I'm probably just a dinosaur who grew up under Python 2.  `pathlib` wasn't in the standard library until 3.6, so I haven't been using it.  I've written a lot of shell and Perl scripts to do filename processing through strings, so the procedural approach of `os.path` makes sense to me.
* I like the idea of `pathlib`
* I like the file-system aware things like `glob`, `exists`, and related.  Pulling in a separate `from glob import glob` has always felt a little funny.  I really like `.suffixes` vs. `.suffix`.  Both are useful and now it's clear what you're doing.  I always had to check `os.path.splitext` to remind myself which it would do.
* I don't like that if I have a `Path` object as the base and then want to add on to it, I have to convert to a string and then back to Path.  It's not a big deal, but it switches me back out of the OO model that `pathlib` was trying to encourage me to think in.
```
from pathlib import Path
foo = Path('a/b/c')
i = 10
Path(f'{foo}_{i}.txt')
```
* I don't like that if I really want a string, the recommendation is to
```
from pathlib import Path
import os
foo = Path('a/b/c')
os.fspath(foo)
```
If the point of `pathlib` is to be an aware object, why do we still need `os.fspath`?
  -- I know the answer.  `Path` does know how to safely convert to a str.  It's a duck-typing safety thing for the caller.  Code that's expecting a `str` or `Path` but relying on `str()` to do the conversion doesn't robustly do the right thing when passed arbitrary objects.  But in practice I still have to `import os; os.fspath()`.
* I really don't like the operator overloading of `A / 'foo.txt'`.  My brain just doesn't read it correctly.  I should likely submit a PR on my brain and join the modern world.

wmwv:
But for the purpose of this Issue, why don't I spend some quality practice time integrating the `pathlib` approach more deeply in my brain.    I'll revert the `os.path` changes and make `get_local_file_list` use the `Path` object glob to practice.

djperrefort:
Closing as stale

PR #25, opened 2019-07-30
tjraen/rapid

troyraen:
@cnmazz, can you test this? I think you'll just need to do the following:

command line:
```bash
pip install -r requirements.txt
```

python:
```python
from broker.classify import rapid as pbr
lst = pbr.collect_ZTF_alerts(max_alerts=1)
res = pbr.classify(lst, plot=True, use_redshift=True)
```


troyraen:
@cnmazz Oh, I should give you a heads up that I do get a lot of warnings, but it still runs. Some are related to my ad hoc use of Rongpu's random forest algorithm, so those will go away when I fix that. I saw that the Travis checks failed, so I'll have to check those out. There may be specific version dependencies that I need to add.

troyraen:
@cnmazz Actually, hold off a bit on testing this. I'm going to restructure the code a little bit to use a `value_added()` function that calls everything else and returns all the value added products. Should be able to finish it today.

cnmazz:
Okay, let me know when it's ready and I'll take a look at it!

troyraen:
I have incorporated the classification into a new value_added branch, so tjraen/rapid does not need to be merged with master.

PR #28, opened 2020-03-02
Adds kafka consumer

djperrefort:
This PR replaces the ``alert_aquisition`` and ``data_upload`` modules with the ``consumer`` module. Some edits are made to the docs, but they need a more dedicated pass through that is beyond this scope.


broker/consumer.py
```diff
@@ -0,0 +1,174 @@
...
+    # 'sasl.kerberos.keytab':
+}
+
+
+class GCSKafkaConsumer(Consumer):
+    """Ingests data from a kafka stream into big_query"""
+
+    def __init__(self, kafka_config, bucket_name, kafka_topic,
+                 pubsub_topic, debug=False):
+        """Ingests data from a kafka stream into big_query
```
wmwv: "And stores a copy in GCS"

broker/consumer.py
```diff
@@ -0,0 +1,174 @@
...
+   config = DEFAULT_ZTF_CONFIG.copy()
+   config['sasl.kerberos.keytab'] = '<Path to authentication file>'
+   config['sasl.kerberos.principal'] = '<>'
+   print(config)
+
+   # Create a consumer
+   c = GCSKafkaConsumer(
+       kafka_config=config,
+       bucket_name='my-GCS-bucket-name',
+       kafka_topics='ztf_20200301_programid1',
```
wmwv: `kafka_topic`  no "s"

broker/consumer.py
```diff
@@ -0,0 +1,174 @@
...
+       pubsub_topic='my-GCS-PubSub-name',
+       debug=True  # Use debug to run without updating your kafka offset
+   )
+
+   # Ingest alerts one at a time indefinitely
+   c.run()
```
wmwv: How do I check the pub sub topic to see these messages?

docs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+sphinx
+sphinx_rtd_theme
```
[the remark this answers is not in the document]
djperrefort: This is the extra requirements file for the docs. I'll add it to the package requirements.

broker/consumer.py
djperrefort: In GCS select `Pub/Sub` -> `Topics` -> `testing-topic`. Programatically I'm not sure but @cnmazz might know.

djperrefort:
I'll merge once @troyraen has a chance to review.

broker/consumer.py
```diff
@@ -0,0 +1,174 @@
...
+   config['sasl.kerberos.principal'] = '<>'
+   print(config)
+
+   # Create a consumer
+   c = GCSKafkaConsumer(
+       kafka_config=config,
+       bucket_name='my-GCS-bucket-name',
+       kafka_topic='ztf_20200301_programid1',
+       pubsub_topic='my-GCS-PubSub-name',
+       debug=True  # Use debug to run without updating your kafka offset
```
troyraen: Naively following these instructions and did not create the bucket first. Got an error on this line `debug=True`:

`BadRequest: 400 GET https://www.googleapis.com/storage/v1/b/my-GCS-bucket-name-Troys-test?projection=noAcl: Invalid bucket name: 'my-GCS-bucket-name-Troys-test'`

Then it kept trying to connect, and returning errors. `ctrl-c` had no effect and I had to shut down my terminal.

I suggest we either add the line of code that creates the bucket (not sure what this is) or just a note saying the bucket needs to exist.

broker/consumer.py
```diff
@@ -0,0 +1,174 @@
...
+   # Define connection configuration using default values as a starting point
+   config = DEFAULT_ZTF_CONFIG.copy()
+   config['sasl.kerberos.keytab'] = '<Path to authentication file>'
+   config['sasl.kerberos.principal'] = '<>'
+   print(config)
+
+   # Create a consumer
+   c = GCSKafkaConsumer(
+       kafka_config=config,
+       bucket_name='my-GCS-bucket-name',
```
troyraen: Bucket name cannot include capital letters

broker/consumer.py
troyraen: The problem forcing me to shut down the terminal was actually caused by not setting `os.environ['KRB5_CONFIG'] = './krb5.conf'` as detailed in the authentication notebook. This module probably assumes one is already authenticated, and in that case the error returned about the bucket not existing is self-explanatory and not fatal, so the code is probably fine the way it is.

broker/consumer.py
djperrefort: Fixed - Thanks @troyraen!

PR #30, opened 2020-03-08
Fixes Travis and introduces usage of PEP 484

djperrefort:
~This PR introduces the usage of PEP 484 to most of the package, and also fixes the Travis/RTD config files so the docs and tests are running again.~

This PR introduces the usage of PEP 484 to most of the package, and also fixes the Travis config files so the tests are running automatically again. I thought I had fixed the ReadTheDocs build, but we are still using too many installation resources and RTD is killing the build. It seems like installing `tensorflow` is the problem. I don't have a clean solutuion for this so I'll think on it more, but the remaining edits still warrant their own PR.


broker/ztf_archive/_download_data.py
```diff
@@ -97,17 +101,17 @@ def _download_alerts_file(file_name, out_dir, block_size, verbose):
             data.extractall(out_dir)
 
 
-def download_data_date(year, month, day, block_size=1024, verbose=True):
+def download_data_date(year, month, day, block_size=1024, verbose=True) -> None:
```
troyraen: Is there a reason these args don't have a type specified? This syntax is new to me, but based on the other functions I think this should be:

```
def download_data_date(
        year: int, 
        month: int, 
        day: int, 
        block_size: int = 1024, 
        verbose: bool = True) -> None:
```

docs/source/index.rst
```diff
@@ -4,7 +4,7 @@ What is Pitt Broker
 ===================
 
 **Pitt Broker** is a cloud-based, alert distribution service designed to provide
```
troyraen: I suggest we change "Pitt Broker" to "Pitt-Google Broker" (4 instances in this file).

djperrefort: Agreed

broker/ztf_archive/_download_data.py
djperrefort: Correct, I must have just missed those.

PR #32, opened 2020-03-14
Adds alert formatting module

djperrefort:
@troyraen here are some thoughts I had on formatting the incoming alert data. I'm still a bit unclear on the exact changes that need to be made to the alerts, so  haven't implemented the actual logic for modifying the schemas.


djperrefort:
This PR has served its purpose as an example and has been moved to `tjraen/alert_formatting`

PR #33, opened 2020-03-20
tjraen/alert formatting

troyraen:


broker/alert_ingestion/consume.py
```diff
@@ -156,8 +189,44 @@ def close(self) -> None:
...
+        """
+
+        # get the corrected schema, if it exists
+        try:
+            f = f'valid_schemas/{survey}_v{version}.pkl'
+            with open(Path(__file__).resolve().parent / f, 'rb') as file:
+                valid_schema = pickle.load(file)
+
+        except FileNotFoundError:
+            return
```
djperrefort: I would expect this to raise a warning. If the alert is not formatted correctly it may or may not raise an error when we upload to BQ (depending on the schema version and survey), but either way it would be better to catch this early in the log. 

`log.error(f'No formatted schema available for {survey} version {version}`)

broker/alert_ingestion/consume.py
```diff
@@ -156,8 +189,44 @@ def close(self) -> None:
...
+        temp_file.seek(0)
+        for r in fastavro.reader(temp_file):
+            data.append(r)
+
+        # write the corrected file
+        temp_file.seek(0)
+        fastavro.writer(temp_file, valid_schema, data)
+        temp_file.truncate() # truncate at current position (removes leftover data)
+
+        return
```
djperrefort: Remove this line.

tests/test_format_alerts.py
```diff
@@ -0,0 +1,113 @@
...
+
+test_alerts_dir = Path(__file__).parent / 'test_alerts'
+
+
+# Quick thoughts on test outlines for alert formatting
+class AlertFormattingDataUnchanged(TestCase):
+    """Test data we retrieve from file before / after formatting is the same"""
+
+    def get_survey_and_schema(self, filename: str) -> (str, float):
+        """ Parses the filename to get the survey and version"""
```
djperrefort: When including test files, consider renaming them with the survey, version, and id for simplicity. For eg., so you can do something like

`survey, version, alert_id = file_path.stem.split('_')`

instead of 

`version = float('.'.join(f[1].strip('v').split('-')))`

tests/test_format_alerts.py
```diff
@@ -0,0 +1,113 @@
...
+                with open(path, 'rb') as f:
+                    temp_file.write(f.read())
+                temp_file.seek(0)
+                consume.GCSKafkaConsumer.fix_schema(temp_file, survey, version)
+
+                temp_file.seek(0)
+                # Get the data after correcting the schema
+                __, corrected_data = _load_Avro(temp_file)
+
+            self.assertEqual(original_data, corrected_data)
```
djperrefort: Add a failure message (third argument position) that includes the survey and version. Otherwise you don't know where in the for loop this fails.

tests/test_format_alerts.py
```diff
@@ -0,0 +1,113 @@
...
+        f = filename.split('_')
+        survey = f[0]
+        version = float('.'.join(f[1].strip('v').split('-')))
+        return (survey, version)
+
+    def test_data_unchanged(self):
+        # Test all alerts in test_alerts_dir
+        for path in test_alerts_dir.glob('*.avro'):
+            # Get the original data
+            __, original_data = _load_Avro(str(path))
```
djperrefort: When using a UnitTest assert in a for loop I often define the for loop content as a dedicated method `def assertDataUnchanged(self, path)`.

tests/test_format_alerts.py
```diff
@@ -0,0 +1,113 @@
...
+    """Test data we retrieve from file before / after formatting is the same"""
+
+    def get_survey_and_schema(self, filename: str) -> (str, float):
+        """ Parses the filename to get the survey and version"""
+        f = filename.split('_')
+        survey = f[0]
+        version = float('.'.join(f[1].strip('v').split('-')))
+        return (survey, version)
+
+    def test_data_unchanged(self):
```
djperrefort: I also fall into the habit of test functions that iterate over conditions, but it is clearer in the test suite output when you write one method to test the condition (e.g., `assertDataUnchanged(self, path)` in a following comment) and additional dedicated methods that call `assertDataUnchanged` for each survey / version.

troyraen:
@djperrefort Will you check this again? I think everything is ready except the guess survey and version functions (still no live streams from ZTF).

tests/test_format_alerts.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+
+"""This file provides tests for the alert formatting performed prior to storage
+in GCS. This includes testing the generation of the valid schema pickle file by
+the ``alert_ingestion.valid_schemas.gen_valid_schema`` module, and the
+correction of alert schemas by the ``alert_ingestion.consume`` module.
+
+Before running these tests:
```
djperrefort: This section might be useful to move into `tests/__init__.py` but I leave it at your discretion.

tests/test_format_alerts.py
```diff
@@ -0,0 +1,213 @@
...
+        survey, version = 'ztf', '3.3'
+        path = test_alert_path[f'{survey}_{version}']
+
+        # Correct the schema and upload to BQ
+        with consume.TempAlertFile(max_size=max_alert_size, mode='r+b') as temp_file:
+            run_consume_fix_schema(path, temp_file)
+            self.upload_to_BigQuery(temp_file)
+
+
+    def upload_to_BigQuery(self, temp_file):
```
djperrefort: may want to rename this to start with the word `assert` so it's clear this is design to potentially fail, error, or pass.

broker/alert_ingestion/valid_schemas/gen_valid_schema.py
```diff
@@ -0,0 +1,195 @@
...
+                schema['fields'][l1]['type'][1]['items']['fields'][l2] = _reverse_types(l2_field)
+
+    # fastavro removes the top level doc item (I don't know why)
+    # add it back in:
+    schema['doc'] = 'avro alert schema for ZTF (www.ztf.caltech.edu)'
+
+    return schema
+
+def _reverse_types(field: dict) -> dict:
+    """ Reverses the order of field['type'] if it is a list _and_ field['default'] is null or is not specified. Otherwise the field is returned unchanged. This is intended to move the 'null' element to the beginning on the type list, but it is up to the user to make sure 'null' is at the end of the list before calling this function.
```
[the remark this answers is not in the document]
troyraen: Oh yes, thank you, that would be better.

broker/alert_ingestion/valid_schemas/gen_valid_schema.py
```diff
@@ -0,0 +1,195 @@
...
+
+        # do the nested candidate fields
+        if l1_field['name'] == 'candidate':
+            for l2, l2_field in enumerate(l1_field['type']['fields']):
+                schema['fields'][l1]['type']['fields'][l2] = _reverse_types(l2_field)
+
+        # do the nested prv_candidate fields
+        if l1_field['name'] == 'prv_candidates':
+            for l2, l2_field in enumerate(l1_field['type'][1]['items']['fields']):
+                schema['fields'][l1]['type'][1]['items']['fields'][l2] = _reverse_types(l2_field)
```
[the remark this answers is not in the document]
troyraen: I agree that it's difficult to follow... I didn't even like it when I wrote it. I had a hard time getting it to work when I defined intermediate variables, and I'm not sure why. Since the schema file is already generated we will never need to use this function again, so I'd prefer not to spend the time to change this and make sure it works.

troyraen:
@djperrefort I made most of those changes. There are 2 that I didn't make to gen_valid_schema.py because those functions have already been used to generate the valid schema file and don't need to be used again. So I didn't want to spend the time to change it.

I made most of the changes suggested by pep8. However, there are a few lines that are too long where it's not obvious to me how to shorten them. I think they're fine (I noticed there are some lines you wrote that are also too long which makes me think this isn't too big of a deal), but let me know if I'm wrong.

broker/alert_ingestion/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+# -*- coding: UTF-8 -*-
+
+"""The ``alert_ingestion`` module handles all logic related to the ingestion
+of Kafka alerts into the Google Cloud Platform. This includes correcting
+schema definitions for compliance with strict Avro standards, uploading the
+alert data to Cloud Storage, and publishing PubSub status messages for the
+ingested alerts.
+"""
+
+from . import consume
```
djperrefort: `consume, gen_valid_schema`

broker/alert_ingestion/gen_valid_schema.py
```diff
@@ -0,0 +1,208 @@
...
+        raise RuntimeError(err_msg)
+
+    else:
+        valid_schema = format_func(schema)  # get the corrected schema
+        _write_dict_to_pickle(valid_schema, fout)
+
+    return valid_schema
+
+
+def _fix_schema_ZTF_v3_3(schema: dict) -> dict:
```
djperrefort:
```
def _fix_schema_ZTF_v3_3(schema: dict) -> dict:
    """ Corrects the ZTF version 3.3 schema to comply with the strict Avro
    validation requirements of BigQuery.
    
    Args:
        schema : Avro schema header, as returned by _load_Avro()
    
    Returns:
        schema : updated schema
    """
    
    for l1, l1_field in enumerate(schema['fields']):  # l1_field is a dict
```

broker/alert_ingestion/gen_valid_schema.py
```diff
@@ -0,0 +1,208 @@
...
+    schema = avro_reader.writer_schema
+    data = [r for r in avro_reader]
+
+    if is_path:
+        f.close()
+
+    return schema, data
+
+
+def _write_dict_to_pickle(schema: dict, fout: str):
```
djperrefort: Add newline after docs here and elsewhere

```
def _write_dict_to_pickle(schema: dict, fout: str):
    """Writes the schema to file as a dictionary"""

    with open(fout, 'wb') as file:
```

broker/alert_ingestion/gen_valid_schema.py
```diff
@@ -0,0 +1,208 @@
...
+
+    for sch, fout in z:
+        schema_bytes = json.dumps(sch).encode('utf-8')
+        with open(fout, 'wb') as f:
+            f.write(schema_bytes)
+
+    return None
+
+
+def _write_Avro(fout: str, schema: dict, data: list):
```
djperrefort: `def _write_Avro(fout: str, schema: dict, data: list) -> None:`

broker/alert_ingestion/gen_valid_schema.py
```diff
@@ -0,0 +1,208 @@
...
+                new_types = field['type']
+        except KeyError:  # default not specified -> reverse the list
+            new_types = field['type'][::-1]
+
+        field['type'] = new_types
+
+    return field
+
+
+def _load_Avro(fin):
```
djperrefort: Include return types in type hints `_load_avro(fin: TextIo) -> ...`

troyraen:
@djperrefort  The Travis build is failing with this error:

```
_________________ ERROR collecting tests/test_format_alerts.py _________________
tests/test_format_alerts.py:31: in <module>
    from broker.alert_ingestion import consume
broker/alert_ingestion/__init__.py:11: in <module>
    from . import consume, gen_valid_schema
broker/alert_ingestion/consume.py:136: in <module>
    class GCSKafkaConsumer(Consumer):
broker/alert_ingestion/consume.py:253: in GCSKafkaConsumer
    def publish_pubsub(self, message: str) -> Future:
E   NameError: name 'Future' is not defined
```

`Future` is imported in `consume.py` with the following:
```
if not os.getenv('GPB_OFFLINE', False):
    from google.cloud import pubsub, storage
    from google.cloud.pubsub_v1.publisher.futures import Future
```

I think the problem is that Travis sets `GPB_OFFLINE=True` and so `Future` is not imported. What's the right way to fix this?

djperrefort:
This is another relic of being unfamiliar with GCP and should be implimented correctly now that we have a better handle on what is going on. Step one is to create a dedicated, API key for travis to use and then upload it to travis following the instructions [here](https://cloud.google.com/solutions/continuous-delivery-with-travis-ci#create_the_books_api_key). Step two is to remove all instances of the `if not os.getenv('GPB_OFFLINE', False):` and rely instead on the travis API key.

troyraen:
Is this something that can be pushed off until later? I really need to finish getting the cloud function done so that I can work on my paper, and this is not a 15 minute task.

troyraen:
The immediate issue is that I can't merge this branch into master because the build is failing. Maybe we should chat when you have a minute.

PR #34, opened 2020-03-30
tjraen/gcs2bq

troyraen:
Adds the Cloud Function that listens to a GCS bucket and uploads new files to a BQ table.

I suggest that we consider moving the `publish_pubsub()` function (in `broker/alert_ingestion/GCS_to_BQ/main.py`) somewhere else so it can be used in functions across the broker.


troyraen:
@djperrefort I think this branch is now ready to merge except for updating the docs. Will you review it?

PR #35, opened 2020-04-24
Cnm tjr/pubsub updates

troyraen:
Updates the broker to use a generic `publish_pubsub()` wrapper function rather than a separate function for each call to Google's `publisher.publish()`.

The `message_service.subscribe_alerts()` function doesn't seem to be working. It fails at `message = pickle.loads(encoded)` with error `_pickle.UnpicklingError: invalid load key, 'O'.` We're not actually using this function right now, so I didn't dig into fixing it. I have skipped the `test_input_match_output()` test that calls it.

### DIFF
--- a/broker/alert_ingestion/GCS_to_BQ/main.py
+++ b/broker/alert_ingestion/GCS_to_BQ/main.py
@@ -105,7 +105,7 @@ def stream_GCS_to_BQ(data: dict, context: dict) -> str:
     # Publish PubSub message if BQ upload was successful
     if error_result is None:
         topic = bucket_resources[bucket_name]['PS_TOPIC']
-        publish_pubsub(topic, file_name)
+        publish_pubsub(topic, file_name.encode('UTF-8'))
 
     return error_result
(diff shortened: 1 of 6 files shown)


troyraen:
Since `future.results()` raises (rather than returns) an exception if the job fails, I'm not sure if my call to `self.assertIs(type(future_result), str)` is the best way to do this test. Open to suggestions, but if you don't have any I think this is good enough for now.

cnmazz:
I'll have to think about whether there's a better way to do the test_publish_pubsub(), but I agree that it's fine for now. I really don't know why subscribe_alerts() isn't working, but I also agree that it's not critical. I can work on figuring this out later, but I do remember using pickle because that was the easiest way I found to handle encoding/decoding dictionaries, so maybe we'll have to be careful with how we encode the alerts so that they can be easily decoded. Otherwise, I think this is good to go!

troyraen:
Thanks! I'll go ahead and merge it in.

PR #37, opened 2020-08-06
Clarifies deployment documentation

djperrefort:
This PR separates the deployment and installation documentation into distinct sections. Various edits are also made for clarity.


djperrefort:
It looks like this is failing because we deleted the testing resources in GCP.

troyraen:
This can be fixed by running `auto_setup()` from `broker.gcp_setup` (Documenting offline conversation)

troyraen:
I also think the `deployment/overview.rst` page should actually list the steps one needs to take to deploy the broker along with links to where (in our documentation) the detailed instructions are. We can come back to this after we redeploy it ourselves.

djperrefort:
As a note to self, I should add to the documentation that most of the docker tasks require `sudo`.

PR #60, opened 2021-05-06
V/0.4.0/tjr

troyraen:
__Major Updates and New Features__

1. Accommodate generic survey: (1) [Renames resources](https://github.com/mwvgroup/Pitt-Google-Broker/tree/v/0.4.0/tjr/version_tracking/v0.4.0/rename-resources.md). (2) Adds a ["survey" keyword](https://github.com/mwvgroup/Pitt-Google-Broker/blob/v/0.4.0/tjr/docs/source/broker/broker-instance-keywords.md). (3) Adds a `schema_maps` module (see below) and configures all components to use it.
2. Creates a [`broker_utils`](https://github.com/mwvgroup/Pitt-Google-Broker/tree/v/0.4.0/tjr/broker/broker_utils/broker_utils) python package. Access with `pip install pgb-broker-utils` and `import broker_utils`. Includes:
- `beam_transforms`: custom transforms used in Beam jobs.
- `consumer_sim`: moved from `dev_utils` and updated. See [new docs](https://github.com/mwvgroup/Pitt-Google-Broker/blob/v/0.4.0/tjr/docs/source/broker/consumer-simulator.md).
- `data_utils`: generally useful functions for dealing with the data (`alert_avro_to_dict`, `mag_to_flux`, etc.).
- `gcp_utils`: some common developer interactions with GCP (load a row to BigQuery, etc.).
- `schema_maps`: retrieve a schema map which can be used to translate field names of a particular survey into generic names used by the broker.
3. Listen for Kafka topics: Consumer VM now polls for available Kafka topics and begins consuming once its assigned topic becomes active.
4. Documentation updates for above changes, and reorganization. READMEs split into multiple files and moved to [docs/source/broker](https://github.com/mwvgroup/Pitt-Google-Broker/tree/v/0.4.0/tjr/docs/source/broker). Full integration with ReadtheDocs is not yet complete, but this is a big step forward.

__Notes__

This PR introduces changes that are backwards incompatible, so perhaps this should end up being tagged `v1.0.0`.

See [version_tracking/v0.4.0](https://github.com/mwvgroup/Pitt-Google-Broker/tree/v/0.4.0/tjr/version_tracking/v0.4.0) for docs on creating and testing this branch, including links to dashboards showing test results.

ZTF broker seems to run like normal, no more or different bugs/errors than normal.

DECAT broker's Dataflow jobs both work (many alerts make it all the way through), but are both very buggy. May not be stable for production use. `bq_sink` has large number of errors, apparently due to `mag=NaN` (see Issue #61). `value_added` producing large number of known issues (for example, [see here](https://github.com/troyraen/PGB_testing/issues/5)).


broker/beam/value_added/value_added.py
```diff
@@ -197,7 +217,15 @@ def run(PROJECTID, sources, sinks, pipeline_args, salt2_configs):
         help="Google Cloud Platform project name.\n",
     )
     parser.add_argument(
-        "--source_PS_ztf",
+        "--SURVEY",
+        help="Survey this broker instance is associated with.\n",
```
wmwv, line 221: I think of "The Broker" has the entire thing.  Is the nomenclature here that "broker instance" is a per-survey thing?

broker/cloud_functions/ps_to_gcs/requirements.txt
```diff
@@ -5,4 +5,5 @@
 
 fastavro
 google-cloud-logging
-google-cloud-storage
+google-cloud-storage==1.38.0
```
wmwv, line 8: Document why the pinned version?

broker/night_conductor/requirements.txt
```diff
@@ -2,13 +2,16 @@ apache-beam[gcp]==2.25.0
 argparse
 astropy==3.2.1
```
wmwv, line 3: Document pinned versions here and lines below.

broker/setup_broker/requirements.txt
```diff
@@ -2,5 +2,5 @@ google-auth
 google-api-core
 google-cloud-bigquery
 google-cloud-logging
-google-cloud-pubsub
-google-cloud-storage
+google-cloud-pubsub==2.2.0
```
wmwv, line 5: Document pinned versions.

version_tracking/v0.4.0/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+# name some things
+consumerVM="${survey}-consumer-${testid}"
+nconductVM="${survey}-night-conductor-${testid}"
+zone="us-central1-a"
+
+# upload credentials
+consumerDir="/home/broker/consumer"
+localDir="/Users/troyraen/Documents/PGB/repo"
+sudo gcloud compute scp "${localDir}/krb5.conf" "${consumerVM}:/etc/krb5.conf" --zone="$zone"
```
wmwv, line 52: Why do these three lines require `sudo`?

wmwv (approved):
Looks great.
1. One actual change would be to document why some of the versions are pinned before you forget.
2. Other comments are things to think about for future naming and nomenclature.
3. What linter do you use?

README.md
```diff
@@ -1,113 +1,30 @@
...
-    - __Stores in__ GCS bucket: [`ztf-sncosmo/salt2/plot_lc`](https://console.cloud.google.com/storage/browser/ardent-cycling-243415_ztf-sncosmo/salt2/plot_lc?pageState=%28%22StorageObjectListTable%22:%28%22f%22:%22%255B%255D%22%29%29&project=ardent-cycling-243415&prefix=&forceOnObjectsSortingFiltering=false) (lightcurve + Salt2 fit, png)
-    - __Publishes to__ PS topics:
-        - [ `ztf_exgalac_trans`](https://console.cloud.google.com/cloudpubsub/topic/detail/ztf_exgalac_trans?project=ardent-cycling-243415) (alerts passing extragalactic transient filter)
-        - [`ztf_salt2`](https://console.cloud.google.com/cloudpubsub/topic/detail/ztf_salt2?project=ardent-cycling-243415) (Salt2 fit params)
-
-5. __Night Conductor__ (orchestrates GCP resources and jobs to run the broker each night)
-    - __Compute Engine VM:__  [`night-conductor`](https://console.cloud.google.com/compute/instancesDetail/zones/us-central1-a/instances/night-conductor?tab=details&project=ardent-cycling-243415)
-
-<!-- fe Broker Architecture -->
+If you run into issues or need assistance, please open an Issue on GitHub or contact troy.raen@pitt.edu.
```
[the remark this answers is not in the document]
troyraen: I made significant changes to this file in PR #62, so I will add this change to that PR.

broker/beam/README.md
```diff
@@ -19,21 +19,16 @@ and [`ztf_bq_sink/beam_ztf_bq_sink.py`](ztf_bq_sink/beam_ztf_bq_sink.py)):
...
-mkdir -p ztf_bq_sink/beam_helpers
-cp beam_helpers/__init__.py beam_helpers/data_utils.py ztf_bq_sink/beam_helpers/.
-cp -r beam_helpers ztf_value_added/.
+source jobs.config "$PROJECT_ID" "$testid" "$beamdir" "$survey"
 
 #-- Start the ztf value-added processing Dataflow job
-cd ztf_value_added
-python3 beam_ztf_value_added.py \
+cd value_added
+python3 value_added.py \
```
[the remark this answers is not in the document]
troyraen: > I prefer commands to be verbs that indicate to the user what will happen when they run it.

Ah, I like that. So this should be something more like "deploy_value_added.py" or to be really verbose "deploy_value_added_dataflow_job.py". Maybe the actual name of the job, "value_added" should change also.

broker/beam/README.md
```diff
@@ -53,13 +50,13 @@ python3 beam_ztf_value_added.py \
             --salt2_SNthresh ${salt2_SNthresh} \
             --salt2_minNdetections ${salt2_minNdetections} \
             --streaming \
-            --update  # use if updating a currently running job; job_name must match current job
-# Use “Ctrl + C” to regain control of the terminal.
-# this will _not_ cancel the job
+            --update  # use only if updating a currently running job
+
+# Once the job starts, use “Ctrl + C” to regain control of the terminal.
```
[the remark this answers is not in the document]
troyraen: I searched the internet for the answer to this months ago, but came up empty. As I recall, the program SHOULD background itself, but that's just not the behavior I get and I couldn't figure out why. I had to implement what felt like an ugly workaround in night-conductor's "start_beam_jobs.sh" script to capture the job ID and then regain control. I think this is worth revisiting at some point.

broker/beam/README.md
```diff
@@ -53,13 +50,13 @@ python3 beam_ztf_value_added.py \
...
 
 # Start the ztf -> BQ job
-cd ${beamdir} && cd ztf_bq_sink
-python beam_ztf_bq_sink.py \
+cd ${beamdir} && cd bq_sink
+python bq_sink.py \
```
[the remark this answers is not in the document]
troyraen: Sounds good. This creates the job on Dataflow and starts it so I think I would choose "deploy_bq_sink.py" (but my definitely of "deploy" might not be strictly correct here).

Will plan to change the name in a future PR.

version_tracking/v0.4.0/README.md
troyraen: I believe it was because the files are copied to directories that are not owned by the user (in this case, me). But actually, this command doesn't work for me anymore. I now have to copy the files to a directory I own (which does not require sudo) and then ssh in and move them.

broker/beam/value_added/value_added.py
troyraen: > I think of "The Broker" has the entire thing.

I also think of it that way... but there is usually (for me) more than one of them.

> Is the nomenclature here that "broker instance" is a per-survey thing?

Partially, yes. Multiple end-to-end brokers can be running in parallel, each ingesting a different survey and/or running different versions of the code. So I needed a way to differentiate them and settled on calling each one a "broker instance" (even recognizing the potential for confusion because of Google's nomenclature of a Compute Engine "instance", aka a VM). 

I refer to "broker instance"s throughout the docs (and try to clarify what it means in the "Broker Instance Keywords" doc), but this might be confusing to people. I am also not necessarily consistent... at times when I feel like there is no confusion I simply refer to "the broker" when I mean "the broker instance". Maybe people expect that multiple instances can run in parallel (or, more likely, they just don't think about running more than one) and so there is no need to differentiate in the first place? Or perhaps "the broker" really already means "the broker instance" and I've just complicated it unnecessarily?

I'm glad you brought this up, and I think it's worth a longer conversation. I wrestled with it, but have never felt like I've come to a good resolution.

troyraen:
> 1. One actual change would be to document why some of the versions are pinned before you forget.

Thanks, I should have and will in the future. This versioning is something I want to go back through and clean up... many of these came from deployment issues that I couldn't quite figure out at the time and I just ended up pinning a bunch of packages to the versions I was was running locally, because it worked for me locally. I'm sure that some are more strict than necessary, and some are probably not necessary at all.

> 3. What linter do you use?

To document our offline conversation: I haven't been using one, but will start. In the near future I plan to refactor all the python files to bring them to a standard code format and run a linter to clean them up. Then I will be consistent going forward.

Issue #68, opened 2021-08-09
pgb-utils throws an error when trying to plot cutouts


troyraen:
When trying to plot cutouts from an Avro file downloaded from Cloud Storage I'm getting the error:

`OSError: No SIMPLE card found, this file does not appear to be a valid FITS file`

Code to reproduce: 

Follow the [Cloud Storage tutorial](https://pitt-broker.readthedocs.io/en/latest/access-data/cloud-storage.html) using pgb-utils v0.2.5.

This worked in the [Colab tutorial](https://github.com/mwvgroup/Pitt-Google-Broker/blob/master/pgb_utils/tutorials/Pitt-Google-Tutorial.ipynb). To double check, I installed pgb-utils v0.1.4 and tried on the same alert packet as before and was able to plot the cutouts.

I think this is an astropy (or related) dependency issue.

troyraen:
Fixed by PR #70

troyraen:
Documenting the work:

Solution summary:
- require astropy==3.2.1
    - fix utils module to accomodate this astropy version

Note the specific error
`OSError: No SIMPLE card found, this file does not appear to be a valid FITS file`
can be avoided by added the kwarg `ignore_missing_simple=True`
to the function `astropy.io.fits.open()`.
However, then we run into the ValueError documented in `pgb_utils.figures.plot_stamp()`,
and fixing this requires astropy 3.2.1.

Work details:

pgb-utils v0.2.5 throws an error when trying to plot cutouts.

Try with colab version of pgb-utils to make sure it still works with a new alert:
```bash
conda create -n pgbutils1 python=3.7
conda activate pgbutils1
pip install pgb-utils==0.1.4

cd /Users/troyraen/Documents/broker/repo2/version_tracking/v0.6.0
python
```

```python
import fastavro
from matplotlib import pyplot as plt
import pgb_utils as pgb

fname = 'ZTF18acegotq.1680242110915010008.ztf_20210808_programid1.avro'
with open(fname, 'rb') as fin:
    alert_list = [r for r in fastavro.reader(fin)]
pgb.figures.plot_lightcurve_cutouts(alert_list[0])
plt.show(block=False)
# this works
```

Find which dependencies make that version work:

```bash
conda create --name issue68 python=3.7
conda activate issue68

pip install ipython
pip install fastavro==1.4.4
pip install matplotlib==3.4.2
# pip install beautifulsoup4==4.8

# pip uninstall astropy
[...]

PR #70, opened 2021-08-12
V/0.6.2/tjr

troyraen:
- [x] Fix Issue #68 by requiring astropy 3.2.1. 
- [x] Update to the utils module to accommodate astropy 3.2.1

@wmwv FYI
@djperrefort there's minor changes here; can you review the PR when you have a minute?


pgb_utils/pgb_utils/figures.py
```diff
@@ -85,14 +85,10 @@ def plot_stamp(stamp, fig=None, subplot=None, **kwargs):
                 subplot = (1,1,1)
             ffig = aplpy.FITSFigure(hdul[0],figure=fig, subplot=subplot, **kwargs)
             ffig.show_grayscale(stretch='arcsinh')
-            # This has been throwing: `ValueError: a must be > 0 and <= 1`
-            # I think I have fixed this by requiring
-            # astropy==3.2.1
-            # astropy-healpix==0.6
-            # beautifulsoup4==4.8
+            # this has been throwing: `ValueError: a must be > 0 and <= 1`
```
djperrefort, line 88: Comments traditionally go before the code they describe. Docstring goes after.

pgb_utils/requirements.txt
```diff
@@ -1,14 +1,7 @@
 aplpy
-astropy>=4.0  # allow create Table from row dicts with different keys
-# astropy==3.2.1
-# astropy-healpix==0.6
-# beautifulsoup4==4.8
+astropy==3.2.1  # required by figures.plot_stamp(), see comments there
```
djperrefort, line 2: I don't think this comment is necessary. Specifically because I don't know at a glance where  `there` is and pointing at the code is just another thing to maintain. I think it's assumed by using `==`` that using other versions might break things.

version_tracking/v0.6/v0.6.2/issue68.md
```diff
@@ -0,0 +1,130 @@
+# Issue \#68
+
+[\#68](https://github.com/mwvgroup/Pitt-Google-Broker/issues/68)
+
+Solution summary:
+- require astropy==3.2.1
+    - fix utils module to accomodate this astropy version
+
+Note the the specific error from issue 68
```
djperrefort, line 9: "the the"

version_tracking/v0.6/v0.6.2/issue68.md
```diff
@@ -0,0 +1,130 @@
+# Issue \#68
```
djperrefort, line 1: I think the contents of this file should be moved to the issue itself.

pgb_utils/pgb_utils/utils.py
```diff
@@ -35,7 +36,19 @@ def alert_dict_to_dataframe(alert_dict: dict) -> pd.DataFrame:
...
+    # astropy 3.2.1 cannot handle dicts with different keys (fixed by v4.1)
+    prv_candidates = []  # recreate with missing keys added
+    for prv_cand in alert_dict["prv_candidates"]:
+        prvc_cols = set(prv_cand.keys())
+        missing_cols = cand_cols - prvc_cols
+        empty = {mc: None for mc in missing_cols}
+        tmp = {**prv_cand, **empty}
+        prv_candidates.append(tmp)
+
+    rows = [candidate] + prv_candidates
```
djperrefort, line 51: This is O(n). Using `.append` would be O(1).

pgb_utils/pgb_utils/utils.py
```diff
@@ -35,7 +36,19 @@ def alert_dict_to_dataframe(alert_dict: dict) -> pd.DataFrame:
 
 def alert_dict_to_table(alert_dict: dict) -> Table:
     """Package a ZTF alert dictionary into an Astopy Table."""
-    rows = [alert_dict["candidate"]] + alert_dict["prv_candidates"]
+    candidate = OrderedDict(alert_dict["candidate"])  # dict
+    cand_cols = set(candidate.keys())
+
+    # astropy 3.2.1 cannot handle dicts with different keys (fixed by v4.1)
+    prv_candidates = []  # recreate with missing keys added
+    for prv_cand in alert_dict["prv_candidates"]:
```
djperrefort, line 44: Every line inside this loop (except the append at the end) is at least O(n). I think this code block is equivalent to 

```# astropy 3.2.1 cannot handle dicts with different keys (fixed by v4.1)
prv_candidates = []  # recreate with missing keys added
for prv_cand in alert_dict["prv_candidates"]:
    for key in candidate:
        prv_cand.setdefault(key, None)

    prv_candidates.append(prv_cand)
```

which only has a single O(n) call.

djperrefort (changes requested):

pgb_utils/requirements.txt
wmwv: I specifically asked @troyraen to document why versions were pinned.  I appreciate the point that `figures.plot_stamp()` doesn't uniquely identify the code, but it does give someone a pointer of what to check for.  Someday we will need to update to AstroPy 4.0 someday for some other function and we need to write down somewhere what to check for breakage when we update.

I think if you say `>=` then you don't really have to say why.  But if you pin a version, then you're specifically saying don't upgrade beyond that because you'll break something.  But you know that that package is going out of date at some point and you know someone will have to updated why.  Pointers, even if it's just a hint, are really key in making sure you can check that you're not going to break the thing that led to pinning the version in the first place.

pgb_utils/pgb_utils/utils.py
wmwv: I haven't read the context here closely ,but I believe `.extend` would be the equivalent operation rather than `.append`.

djperrefort: Ah, I hadn't appreciated that the `candidate` element needed to be the first in the list.

djperrefort: I was thinking of `prv_candidates.append(candidate)`

wmwv: @djperrefort are you advocating for creating a temporary array and then extending that?
```
candidate_tmp = [candidate]
rows = candidate_tmp.extend(prv_candidate)
```

djperrefort: I'm advocating that the list in the above accumulation pattern with `candidate` already in it. I.e., `prv_candidates = [candidate]` before the for loop.

djperrefort: I agree - optimizing this call won't yield a measurable performance gain. I'm mostly pushing on this from a readability point of view. I probably should have merged this comment with my comment on the above for loop.

version_tracking/v0.6/v0.6.2/issue68.md
troyraen: I copied the file contents to the issue comments.

pgb_utils/pgb_utils/utils.py
troyraen: Thanks for this, it's much cleaner. My only issue with it is that it changes the original alert_dict object. How about using `get()` instead of `setdefault()`?

```python
prv_candidates = []  # recreate with missing keys added
for prv_cand in alert_dict["prv_candidates"]:
    for key in candidate:
        prv_cand.setdefault(key, None)

    prv_candidates.append(prv_cand)
```

djperrefort: I'm not clear how using `get` solves the problem. how about using `prv_cand = prv_cand.copy()`?

troyraen: Thanks for this, it's much cleaner. My only issue with it is that it changes the original alert_dict object. How about using `get()` instead of `setdefault()`?

```python
prv_candidates = []  # recreate with missing keys added
for prv_cand in alert_dict["prv_candidates"]:
    prv_cand_tmp = {key: prv_cand.get(key, None) for key in candidate.keys()}
    prv_candidates.append(prv_cand_tmp)
```

troyraen: Sorry, my first comment just copied your code instead of showing mine... 
See my proposed code above. Doesn't that solve the problem?

pgb_utils/pgb_utils/utils.py
troyraen: I updated the code to instantiate the list with `candidate` already in it.
Thanks for the pointers.

troyraen:
@djperrefort I think I've made all requested changes.

djperrefort (approved):

PR #71, opened 2021-08-20
V/0.7.0/tjr

troyraen:
Adds SuperNNova classifier

- [x] Add SuperNNova classifier as a Cloud Function.
- [x] Add the extragalactic transients filter* as a Cloud Function, emitting the Pub/Sub stream that SuperNNova listens to.
- [x] Update broker_utils with tools used to develop the cloud functions.
- [x] Update docs

See especially:
- version_tracking/v0.7.0/README.md
- broker/cloud_functions/classify_snn/main.py
- broker/cloud_functions/filter_exgalac_trans/main.py

This really should have been 3 separate PRs, one for each cloud fnc and one for broker_utils. I am sorry there are so many changes here.


broker/broker_utils/broker_utils/data_utils.py
```diff
@@ -6,24 +6,91 @@
...
+                      If `return_as='df'` the cutouts are always dropped.
+
+    Returns:
+        alert packet in requested format
+    """
+    alert_dict = alert_avro_to_dict(alert_avro)
+
+    if return_as == "dict":
+        if drop_cutouts:
+            return _drop_cutouts(alert_dict, schema_map)
```
troyraen, line 42: @djperrefort Codacy complains about this line, saying:

> "Argument 2 to "_drop_cutouts" has incompatible type "Optional[Dict[Any, Any]]"; expected "Dict[Any, Any]" (error)".

I doesn't seem to know about `typing.Optional` Any ideas on how to fix that? I tried a Google search, but didn't get very far.

Apparently the error is too minor to be flagged here, but I see it on the [Codacy GUI](https://app.codacy.com/gh/mwvgroup/Pitt-Google-Broker/pullRequest?_ga=2.222584475.1032047264.1630857686-1452767873.1630601139)

djperrefort: @troyraen The error message is correct. The inspection tool is telling you that the function signatures are not compatible.

The `_drop_cutouts` function is type hinted to imply that the second argument (`schema_map`) is **not** optional. However,  the signature of the `decode_alert` function indicates that `schema_map` may be `None`. 

Type hinting is used to help prevent bugs that can creep into code due to duck-typing. In this case, either the type hints are incorrect, or the the functions are being called in a way that violates the way the function is supposed to be called.

broker/broker_utils/broker_utils/data_utils.py
```diff
@@ -33,16 +100,43 @@ def alert_dict_to_dataframe(alert_dict: dict, schema_map: dict) -> pd.DataFrame:
     """ Packages an alert into a dataframe.
     Adapted from: https://github.com/ZwickyTransientFacility/ztf-avro-alert/blob/master/notebooks/Filtering_alerts.ipynb
     """
+    if not isinstance(schema_map, dict):
+        raise ValueError("`schema_map` is not a valid dictionary.")
+
     dfc = pd.DataFrame(alert_dict[schema_map['source']], index=[0])
     df_prv = pd.DataFrame(alert_dict[schema_map['prvSources']])
```
cmazzdaher, line 107: Super minor thing, but "df_prv" has an underscore while the other two variables "dfc" and "dflc" don't.

broker/broker_utils/broker_utils/data_utils.py
troyraen: @djperrefort Ah, thank you. Can you help me code this correctly? `schema_map` is __not__ required if `drop_cutouts=False`. So I want it to be optional in `decode_alert()`, even though it is required in `_drop_cutouts()`. How do I write that correctly?

broker/broker_utils/broker_utils/data_utils.py
troyraen: Thanks, and all those names are pretty bad to begin with. I took the opportunity to rename them all, hopefully it's a bit better.

broker/broker_utils/broker_utils/data_utils.py
djperrefort: The simplest way is to LBYL and check the `schema_map` argument is not `None` before calling `_drop_cutouts`. If it is `None` then raise an error.

The more complex way is to add multiple type hint signatures:

```python

from typing import overload

@overload
def decode_alert(
    alert_avro: Union[str, bytes],
    return_as: str,
    schema_map: dict,
    drop_cutouts: bool,
) -> Union[dict, pd.DataFrame]:
    ...


@overload
def decode_alert(
    alert_avro: Union[str, bytes],
    return_as: str,
    schema_map: None,
    drop_cutouts: bool,
) -> Union[dict, pd.DataFrame]:
    ...


def decode_alert(alert_avro, return_as='dict', schema_map=None, drop_cutouts=False):
    ...  # source code here
```

Which one you use depends on how important the type inspections are to your use case. I suggest using the former for internal code and the latter for code you expect to ship out as an API for use by end users.

troyraen:
@cmazzdaher @djperrefort The feedback from Codacy is helpful, but there are a lot of little things to fix. I'd like to get the currently open PRs merged together, and then do a separate PR dedicated to fixing all this stuff. So I'd like to punt on all these Codacy issues for now. None of them look urgent to me.

broker/broker_utils/broker_utils/data_utils.py
troyraen: @djperrefort I added these LBYL checks as you suggested, but it seems overly complicated to me. I have to do 2 checks here, but the function that **actually uses** the `schema_map` already checks the type and raises an error if it's not as expected. 
In other words, this adds 2 redundant checks, requiring 6 new lines of code.  Is this really how I should do it? 

There are other similar situations in the code, and it seems dumb to do redundant checks in every single place. Is there not a good way to let the function that actually uses the `schema_map` handle the type check by itself?

djperrefort: > Is there not a good way to let the function that actually uses the schema_map handle the type check by itself?

I'm not sure I understand the question. That should be what the overload example above is doing.

At the end of the day it's just a type check. It can be ignored and revisited as part of our gradual efforts to address codacy inspections in the future.

troyraen: > That should be what the overload example above is doing.

No, `decode_alert()` is not the function that actually uses the `schema_map`. 
`drop_cutouts()` uses the `schema_map`, and it **already** checks to make sure that it is a valid type. 

> I'm not sure I understand the question.

Is there a best-practice way to allow functions that do not actually use the `schema_map` (they just pass it on to another function) to be ignorant of the type? It seems so much simpler to let them pass `schema_map=None` because `drop_cutouts` will handle the check.

> At the end of the day it's just a type check. It can be ignored and revisited as part of our gradual efforts to address codacy inspections in the future.

Yes, I don't really care about these 6 lines. Really, I'm thinking about all of my code going forward and trying to understand best practices.

troyraen:
I added a couple of logging statements and redeployed the Cloud Function and it appears to be uploading to BigQuery again as expected. I didn't change any of the code except to add the logging statements, so I don't know what was wrong before.

broker/broker_utils/broker_utils/data_utils.py
troyraen: I have since realized that it's probably better to take `schema_map` out of the `decode_alert` signature, and replace it with `**kwargs` which can then be passed on to `_drop_cutouts`.

broker/cloud_functions/filter_exgalac_trans/main.py
```diff
@@ -0,0 +1,117 @@
...
+
+# GCP resources used in this module
+ps_topic = f"{SURVEY}-exgalac_trans_cf"
+if TESTID != "False":  # attach the testid to the names
+    ps_topic = f"{ps_topic}-{TESTID}"
+
+schema_map = schema_maps.load_schema_map(SURVEY, TESTID)
+
+
+def run(msg: dict, context) -> None:
```
wmwv, line 32: Is the deal here that `run` must accept an argument `context`?

broker/cloud_functions/filter_exgalac_trans/main.py
```diff
@@ -0,0 +1,117 @@
...
+        is_positive_sub = candidate["isdiffpos"] == "t"
+
+        if (candidate["distpsnr1"] is None) or (candidate["distpsnr1"] > 1.5):  # arcsec
+            no_pointsource_counterpart = True
+            # closest candidate == star < 1.5 arcsec away -> candidate probably star
+        else:
+            if candidate["sgscore1"] < 0.5:
+                no_pointsource_counterpart = True
+            else:
+                no_pointsource_counterpart = False
```
wmwv, line 93: Yeah, I'll agree with Codacy here that
```
no_pointsource_counterpart = candidate["sgsscore1"] < 0.5
```
Is both more compact and easier to read.

wmwv (approved):
LGTM.

A bunch of small little comments to fix up in the future.  You probably are aware of most of these already.

broker/cloud_functions/classify_snn/ZTF_DMAM_V19_NoC_SNIa_vs_CC_forFink/cli_args.json
```diff
@@ -0,0 +1,260 @@
...
+        "data": false,
+        "data_fraction": 1.0,
+        "data_plasticc_test": false,
+        "data_plasticc_train": false,
+        "data_testing": false,
+        "data_training": false,
+        "debug": false,
+        "done_file": null,
+        "dropout": 0.05,
+        "dump_dir": "/scratch/midway2/amoller/ZTF/dump_DMAM_V19_NoC_SNIa_vs_CC",
```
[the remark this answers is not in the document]
troyraen: This is a file produced during the training of the model. `supernnova` parses it to find some settings.

broker/cloud_functions/classify_snn/ZTF_DMAM_V19_NoC_SNIa_vs_CC_forFink/data_norm.json
```diff
@@ -0,0 +1,27 @@
+{
```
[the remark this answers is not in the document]
troyraen: Not specifically, but I added a README documenting what I do know about the trained model in this directory.

broker/cloud_functions/filter_exgalac_trans/main.py
troyraen: Yes that's right. This function is triggered by a Pub/Sub message via Cloud Functions, which expects both arguments. I added documentation.

broker/broker_utils/broker_utils/schema_maps.py
```diff
@@ -23,6 +25,7 @@ def load_schema_map(survey: str, testid: str, schema: Optional[str] =None) -> di
         testid: Name of the testid associated with the broker instance.
                 Along with the `survey`, this determines which GCS bucket the
                 schema map will be loaded from.
+                If this is a bool, it must be False (indicates a production instance).
```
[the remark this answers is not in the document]
troyraen: Yes, `None` would be better. It needs to be chaned throughout the code (Issue #97).

broker/broker_utils/broker_utils/schema_maps.py
```diff
@@ -38,7 +41,7 @@ def load_schema_map(survey: str, testid: str, schema: Optional[str] =None) -> di
     return schema_map
 
 def _broker_bucket_name(survey, testid):
-    if testid == "False":
+    if testid in ["False", False]:
```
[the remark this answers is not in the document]
troyraen: Yes, this is messy and it is a problem that exists in multiple places that `testid` is used. Issue #97 should fix this.

PR #72, opened 2021-08-23
V/0.7.1/tjr

troyraen:
Starts a system to track metadata for provenance and performance benchmarking.

- [x] Add custom metadata attributes to messages in all Pub/Sub streams, except "alerts".
- [x] Add a script to Night Conductor that processes the Pub/Sub streams each morning, extracts the metadata, and stores it in BigQuery.
- [x] Update broker_utils with tools to do the above.
- [x] Update docs

See especially:
- docs/source/broker/components/tracking-metadata.rst
- broker/night_conductor/end_night/process_pubsub_counters.py
- version_tracking/v0.7/v0.7.1/README.md


wmwv:
If this PR is meant to be changes since `dc1ff8b`, can you change it so that the comparison is against that version?  It's currently against `master`, which means I have to wade through all of the other changes as well.  It's hard for me to figure out what is the changes just against `dc1ff8b`.

troyraen:
@wmwv Done; thanks for the prompt... still trying to learn the best way to organize things for review by others.

broker/broker_utils/broker_utils/beam_transforms.py
```diff
@@ -112,11 +112,23 @@ def process_ztf(self, alertDict):
...
-        Can be updated to return a :class:`~PubsubMessage` object.
-        In that case, change WriteToPubSub() kwarg 'with_attributes' to `True`.
-        See https://beam.apache.org/releases/pydoc/2.26.0/apache_beam.io.external.gcp.pubsub.html?highlight=writetopubsub#apache_beam.io.external.gcp.pubsub.WriteToPubSub
+
+        Returns a `PubsubMessage` object with custom attributes attached.
         """
+        from apache_beam.io.gcp.pubsub import PubsubMessage
         import json
+
+        # collect message attributes
```
wmwv, line 121: Can you add some context about why sometimes we want the value pointed to by `alertDict['alert']` but if that key doesn't exist we want the full `alertDict`.  Reading this out of context, that sounds like an odd way for the alert data to be organized.

broker/broker_utils/broker_utils/gcp_utils.py
```diff
@@ -85,9 +89,19 @@ def pull(
...
-                    If None, the environment variable GOOGLE_CLOUD_PROJECT will be used.
+                    If None, the module's `pgb_project_id` will be used.
+
+        msg_only: Whether to work with and return the message contents only
+                  or the full packet.
+                  If `return_count` is True, it supersedes the returned object.
 
-        msg_only: Whether to return the message contents only or the full packet.
+        callback: Function used to process each message.
+                  It's input type is determined by the value of `msg_only`.
```
wmwv, line 99: "Its"

broker/broker_utils/requirements.txt
```diff
@@ -4,9 +4,13 @@ fastavro
 google-cloud-bigquery<2.10
 # apache-beam 2.25.0 requires pyarrow<0.18.0,>=0.15.1
 # but google-cloud-bigquery>=2.10.0 requires pyarrow>=3
+google-cloud-logging
 google-cloud-storage>=1.38
 google-cloud-pubsub<2  # beam requires <2
-numpy
+numpy==1.19.5
+# older pyarrow versions require numpy<1.20 to successfully convert datatypes
```
wmwv, line 11: Put comments before the code line, in this case `numpy=1.19.5`

wmwv: Do you know specifically which older pyarrow versions?

broker/broker_utils/setup.py
```diff
@@ -30,8 +30,8 @@
     requirements = f.read().splitlines()
 
 setup(
-    name='pgb_broker_utils',  # Required
-    version='0.2.10',  # Required
+    name='pgb_broker_utils',
+    version='0.2.20',
```
wmwv, line 34: Is this key line where the version of the broker is tracked?  Consider doing some reading into how people handle this in general.  It seems like having some more visible file that defines version and then is imported into things that need the version might be a way to go here.

wmwv: But also, I'm starting to wonder, if this is specifically the `pgb_broker_utils` version, that implies that we are tracking versions separately.  If we're going to have separate versions, should these be in separate repos?

broker/night_conductor/end_night/process_pubsub_counters.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+if __name__ == "__main__":  # noqa
+    parser = argparse.ArgumentParser()
+    # survey
+    parser.add_argument(
+        "--survey",
+        dest="survey",
+        default="ztf",
+        help="The broker instance's `survey` keyword. Names the resources.\n",
```
wmwv, line 360: "Broker instance `survey` keyword.  Used to label resources."

Might be more clear.  (similar below for `--testid`).

broker/night_conductor/end_night/process_pubsub_counters.py
```diff
@@ -0,0 +1,386 @@
...
+        "--testid",  # set testid = <mytestid>
+        dest="testid",
+        default="test",
+        help="The broker instance's `testid` keyword. Names the resources\n",
+    )
+    parser.add_argument(
+        "--production",  # set testid = False
+        dest="testid",
+        action="store_false",
+        default="test",
```
wmwv, line 374: If `dest="testid"` then don't respecify the default parameter again.

broker/night_conductor/end_night/process_pubsub_counters.py
```diff
@@ -0,0 +1,386 @@
...
+        help="Sets `testid=False`. Use for broker production instances.\n",
```
wmwv, line 375: I'm confused.  Is `testid` a boolean (implied here) or a string (implied by this `--production` argument).

broker/night_conductor/end_night/process_pubsub_counters.py
```diff
@@ -0,0 +1,386 @@
...
+    )
+    parser.add_argument(
+        "--batch_size",
+        dest="batch_size",
+        default=500,
+        help="Maximum number of Pub/Sub messages to pull per call.",
+    )
+
+    known_args, __ = parser.parse_known_args()
```
wmwv, line 384: Single `_` is fine and the usage I'm familiar with.

broker/night_conductor/end_night/process_pubsub_counters.py
```diff
@@ -0,0 +1,386 @@
...
+
+    run(known_args.survey, known_args.testid, int(known_args.batch_size))
```
wmwv, line 386: Fix the type to enforce `int` in the `--batch_size` `add_argument` string.

broker/night_conductor/requirements.txt
```diff
@@ -2,14 +2,19 @@ apache-beam[gcp]==2.25.0
 argparse
 astropy==3.2.1
 fastavro==0.22.3
-google-cloud-bigquery
+google-cloud-bigquery<2.10
+# apache-beam 2.25.0 requires pyarrow<0.18.0,>=0.15.1
```
wmwv, line 6: Comment before code.

broker/night_conductor/requirements.txt
```diff
@@ -2,14 +2,19 @@ apache-beam[gcp]==2.25.0
...
+# but google-cloud-bigquery>=2.10.0 requires pyarrow>=3
 google-cloud-core>==1.4.1
 google-cloud-dataflow
 google-cloud-datastore>==1.15
 google-cloud-storage==1.38.0
 iminuit==1.4.9
 matplotlib==3.1.1
-numpy==1.18.2
+numpy==1.19.5
+# older pyarrow versions require numpy<1.20 to successfully convert datatypes
```
wmwv, line 15: Comment before code.

wmwv (approved):
This looks like a good start on adding tracking to the system.

1. This repo contains several different modules.  Each has a separate requirements.txt and potentially version information.  Can you outline in part of the design description why this choice?  Having separate repos isn't obviously better, but I would suggest that the repo just has one global version. 
2. How is the Broker code version inserted into the message for tracking?
3. When a classification is run, where is that code version information stored?

broker/broker_utils/requirements.txt
troyraen: I don't know, and I didn't see the answer anywhere in the official Issue. My best guess is pyarrow<1.0.0 (versions 1.0.0 to 5.0.0 were all developed in less than 1 year).

broker/broker_utils/setup.py
troyraen: > It seems like having some more visible file that defines version and then is imported into things that need the version might be a way to go here.

Makes sense, and I have seen others do something similar. I'll look at other examples more closely.

> But also, I'm starting to wonder, if this is specifically the `pgb_broker_utils` version, that implies that we are tracking versions separately. 

Correct, and see my (long) comment below for more discussion.

> If we're going to have separate versions, should these be in separate repos?

Addressed below.

broker/night_conductor/end_night/process_pubsub_counters.py
troyraen: This is an issue throughout the broker because the shell scripts always pass the `testid` as a string, but in the case of `testid=False` it's more natural for the python scripts to handle it as a bool. I have done some "hotfixes" to handle this, but I should clean them up for clarity.

troyraen:
Addressing various comments about different modules/versions:

> 1. This repo contains several different modules.  Each has a separate requirements.txt and potentially version information.  Can you outline in part of the design description why this choice? Having separate repos isn't obviously better, but I would suggest that the repo just has one global version.

Yes, I will work on cleaning this up and clarifying the documentation. Let's make some decisions first...

> 2. How is the Broker code version inserted into the message for tracking?
> 3. When a classification is run, where is that code version information stored?

Currently, neither happen. The tracking system implemented in this PR does not yet track the code versions. Now that the system is set up, these can be easily added once we make some decisions on the issues you bring up in 1.

To this end, here are some descriptions of the current status and some thoughts.

Modules/packages in this repo:

1. "The broker" (i.e., everything under the broker dir): Currently, the only "version" label that gets assigned is the tagged release on GitHub. There is no version number that is currently tracked in the code itself, but I agree that there should be. I'd like to tie it to the `pgb-broker-utils` version, but I'm not yet sure whether this makes sense.

2. `pgb-broker-utils` (i.e., the broker/broker_utils dir): A collection of tools used by multiple broker components. It is `pip`-installable, and therefore has official versions. It does not depend on the rest of the broker code and could be split off into its own repo. But since the broker relies heavily on it, I'd advocate for leaving it in this repo but un-nesting it from the broker dir for clarity.

3. `pgb-utils` (i.e., the pgb_utils dir): A collection of tools developed for the tutorials, intended for use by external users. It is `pip`-installable, and therefore has official versions. It probably makes sense to split this off into its own repo (and this would match what the other brokers do). 

4. Top-level `requirements.txt`: For use by developers during initial setup. This can probably be dropped; developers should just install `pgb-broker-utils`.

5. Top-level `setup.py` file: This is required by Read The Docs, and only used for that purpose. I'll see if I can move it to the docs dir. (There is a nominal "version" defined in there, but I never update it.)

5. Various `requirements.txt` files in Cloud Functions (dirs under broker/cloud_functions) and in /broker/night_conductor. These are used by GCP to deploy/install the individual components. In the case of Cloud Functions, these files are *required* to be in these locations. I can improve the documentation for clarity.

troyraen:
@wmwv To extract my main question from my long comment above:

Do you think that "The broker" and `pgb-broker-utils` can share a version number? 

The conflict I see is, what happens when "The broker" code is updated, but `pgb-broker-utils` is not? Do we push a `pgb-broker-utils` update to PyPI where the only change is the version number? Or maybe we don't, and the published package skips versions that don't affect it?

wmwv:
If `pgb-broker-utils` and `pgp-utils` are separately pip-installable and thus are externally independent entities and recognized with their own version numbers, then it sounds like the decision is already made to have separate versions for the different pieces.  I agree that organizing things so that each separately versioned part is in its own separate directory within the repo is a good step. 

Separating out tutorials and user-tools would also be a reasonable choice, but we'd still have at least two different packages in the current repo.  So either's fine.

wmwv:
I'd be interested in learning more about how some other repos structure things.  Let me know what you find out as a best/reasonable practice for doing this.

wmwv:
And go ahead a merge this in first.  I don't mean the versioning question to hold up this PR.

Issue #82, opened 2021-09-02
Integration Testing


troyraen:
Initial thoughts towards automated integration testing :
- GitHub Action
- Build a testing instance of the broker (run the setup script)
- Push some alerts through using the consumer simulator
- Check the results using tools in either `broker-utils` or `pgb-utils`. Query BigQuery tables, pull Pub/Sub streams, download avro files (plot cutouts).

If it's helpful, there is a VM called stream-looper that is always on and already has the consumer simulator installed. (It publishes the ztf-loop Pub/Sub stream). It can easily be utilized to run consumer simulators for testing broker instances. I sometimes do this when I run these tests manually.

troyraen:
@djperrefort Here is the python code to load a single alert from file and publish it to pubsub

```python
from broker_utils import data_utils, gcp_utils, schema_maps

# broker instance keywords
survey = 'ztf'
testid = 'integrationtest'

# load alert from file to dict
# here's a convenient test file that's in the repo
fname = 'version_tracking/v0.6.0/ZTF18acegotq.1680242110915010008.ztf_20210808_programid1.avro'
# avoid a bug; Issue #88
drop_cutouts = True
schema_map = schema_maps.load_schema_map(survey, testid)  # needed for drop_cutouts=True
alert_dict = data_utils.decode_alert(
    fname, return_as='dict', drop_cutouts=drop_cutouts, schema_map=schema_map
)

# publish to Pub/Sub
# topic = f'{survey}-alerts-{testid}'  # use this for integration testing
topic = 'test'  # I use this topic for random tests
gcp_utils.publish_pubsub(topic, alert_dict)
```

If you need to load a `schema_map` but haven't set up a testing instance yet, just set `testid=False` in that call. That will load the schema map from the production instance.

troyraen:
I will collect some alerts that I know will make it through all the filters/classifiers/fitters, plus some that will not, so that we have a good set to test from.

@djperrefort should I add them somewhere in the issues/integration_workflow branch you created?

[...]

Issue #79, opened 2021-09-02
Organize and track broker versions


troyraen:
Goals: 
- Design a system to assign a global version for the code in this repo.
- Collect the broker version, and versions of other important packages, in the [metadata tracking system](https://pitt-broker.readthedocs.io/en/latest/broker/components/tracking-metadata.html).

To do:
- [ ] Create a dedicated file to hold the broker version, and a method for other files to retrieve the version.
- [ ] Create and document the process by which the version gets bumped. Perhaps this should be handled by a GitHub Action, triggered by a merge into master.
- [ ] Configure broker components to collect relevant versions and add them to Pub/Sub messages as metadata attributes. The broker's global version may only need to be collected once (i.e., by a single component). Other components need to collect versions of packages important for tracking provenance (e.g., SuperNNova component needs to collect the `supernnova` version).
- [ ] Complete #78 so that the `pgb-utils` version doesn't conflict/cause confusion with the global repo version.

Open question:
Should `broker-utils` be moved into its own repo? Most broker components use these tools. Since each component is deployed to a separate location in GCP, the `broker-utils` module is `pip`-installable so that we don't have to copy it separately into each component before deployment. This means that it has its own official version, which will not necessarily match the global version for the repo.

This Issue was prompted by the discussion in PR #72.

troyraen:
Re: the "open question" above. 

My initial feeling was that `broker-utils` is such an integral part of "the broker" that I really wanted to keep it in the same repo and figure out how to resolve the version conflict. I have previously looked at other brokers' repos and seen that they have their **user** tools in a separate repo, but I hadn't seen anything comparable to `broker-utils` in a separate repo.

Upon looking more closely:
- Fink has all their science modules in a dedicated repo, separate from the main broker repo. 
- ALeRCE doesn't even seem to have one main "broker" repo. There are 49 different repos at [github.com/alercebroker](https://github.com/alercebroker) and many of them are individual broker components/tools. There is one repo that contains a "framework (...) to create a dockerized pipeline to process an alert stream". So I assume this is how they collect/package/deploy the broker.

Ok, so I'm convinced that just because `broker-utils` is integral to the rest of the "the broker", it doesn't need to be in the same repo. Moving it probably makes the most sense.

(pinging @wmwv)

wmwv:
[...]

Issue #76, opened 2021-09-01
Use TNS to add the IAU name and other info to the alert packets


troyraen:
Add the IAU name to the alert packets, if available. Look these up from TNS. 

Workflow (preliminary thoughts): 
- [ ] Download a CSV of objects nightly. See the TNS page [Daily CSV Staging](https://www.wis-tns.org/content/tns-getting-started#Daily%20CSV%20staging)
- [ ] Load these to BigQuery
- [ ] Do a positional cross match for each alert to identify known objects. Grab and add the info to the alert packet.
- [ ] Bonus: figure out method to store a TNS/IAU identifier, and then use it to cross match future alerts representing known objects. This should be less expensive than repeating the position cross match each time.

PR #85, opened 2021-09-02
Add unittesting workflow

### LINKED ISSUE #73: Disable checking of PRs that always fails.
I believe the current status is that the automatic test that GitHub runs for a PR always fails.  Let's disable that check until we figure out how to change things so that we expect it to succeed.  I'd like to not get into the a bit of expecting the test to fail and ignoring red X marks.

djperrefort:
Relevant to Issues #81 and #82
Closes #73 

- Removes tests for code that doesn't exist anymore
- Adds a github actions workflow for running unittests
- Adds an example unit test (just so the workflow has something to run)
- Deletes travis config file


djperrefort:
@troyraen It looks like there is a dependency issue with the requirements file. I get a similar error when building locally. Are you able to install the broker package in a clean environment?

troyraen:
> I never need to install this all in the same place

Ok, yes, I do have it all installed locally (i.e., in the same place) and I test all the pieces locally. But my point is

> the end-to-end broker will not run from a single machine.

So I've never had a need to install the entire broker "package" together in a clean environment. I install each piece component in the specific clean environment in which it is designed to run.

djperrefort:
The problem is coming from the requirements file in root. The call to `pip install -r requirements.txt` fails with one of two error messages. In Python 3.7 I get the following during the installation of `google-auth`

> INFO: This is taking longer than usual. You might need to provide the dependency resolver with stricter constraints to reduce runtime. If you want to abort this run, you can press Ctrl + C to do so. To improve how pip performs, tell us what happened here: https://pip.pypa.io/surveys/backtracking

In Python 3.8 and 3.9 I get 

> ERROR: Could not find a version that satisfies the requirement astropy==3.2.1

troyraen:
> The problem is coming from the requirements file in root. 

Documenting our offline discussion, try using the file at broker/broker_utils/requirements.txt. The requirements file in root is incomplete and not intended for this use case. I am sorry it's even there to begin with, since its very existence implies that you should be able to use it in the way you are trying to use it. This will get cleaned up as part of #79.

> In Python 3.8 and 3.9 I get
> 
> > ERROR: Could not find a version that satisfies the requirement astropy==3.2.1

I think we need to stick to Python 3.7. You are the 3rd person to report trouble when trying to install a broker component in something other than Python 3.7. For example, Issue #76 is related.

PR #100, opened 2021-09-20
Cross match with AllWISE

troyraen:
Cross match alerts with AllWISE catalog in BigQuery public datasets.
Cloud Run implementation.


troyraen:
I ran this for 2 days, but it appears to be doing a full table scan for each match (offline conversation with @wmwv). It is expensive, and I have disabled it until I can write it more efficiently. Here are some figures:

AllWISE-Trigger-ztf-20210927

GCPbilling-20210927

Billing: 
- Orange bars on Sept 25 and 26 are the DB queries (“Analysis” $169.92). 
- This uses the Pacific timezone, so an observing night gets split across 2 billing days. What you see here is basically 1.5 nights; the second half of last night will show up when Sept 27 billing gets reported.

troyraen:
For reference, here’s the SQL statement and some notes/questions about it.

```
sql_stmnt = f"""
    CREATE TEMP FUNCTION
        IsWithinMaxSep(p1 GEOGRAPHY, p2 GEOGRAPHY, max_sep_arcsec FLOAT64)
            AS (ST_DISTANCE(p1, p2) < max_sep_arcsec * 30.8874796235);
    SELECT {', '.join(cols)}
    FROM `bigquery-public-data.wise_all_sky_data_release.all_wise`
    WHERE
        IsWithinMaxSep(point, ST_GEOGPOINT(@ra, @dec), @max_sep)
        AND ST_CONTAINS(ST_GEOGFROMTEXT(@polygon), point)
    ORDER BY ST_DISTANCE(point, ST_GEOGPOINT(@ra, @dec))
    LIMIT @limit;
"""
```

- ST_DISTANCE() returns the distance (meters) on a sphere with Earth’s radius. 30.8874796235 = meters/arcsec on that sphere
- I understood ST_CONTAINS to restrict the region that was searched, but I guess I’m not sure whether it affects how much data is actually being processed. 
- The all_wise table is clustered by HTM Level 7 and HTM Level 20 index keys.

I don’t know whether the last 2 items play into each other.

wmwv:
Can you give some examples of `@polygon`?

wmwv:
Re-reading
https://cloud.google.com/blog/products/data-analytics/querying-the-stars-with-bigquery-gis
I see that the reported per-point performance is 4 seconds.  So I suspect that the observed time performance is consistent with that in the blog post.

1. I'm still surprised at the resource performance.
2. However, I think we can, and we certainly need to do much better.

troyraen:
@wmwv Unfortunately, we cannot "just calculate" the HTM indexes. The current implementation (at [http://www.skyserver.org/htm/](http://www.skyserver.org/htm/)) is in C#. I can't find any python wrappers. We could (probably) write our own.

An alternative is [HEALPix](https://healpix.jpl.nasa.gov/), which has two python implementations: [`healpy`](https://healpy.readthedocs.io/en/latest/) (original); [`astropy-healpix`](https://astropy-healpix.readthedocs.io/en/latest/index.html) (still in early dev).

[Berriman, et al. 2018](https://arxiv.org/abs/1806.08866) did a comparison and found that the choice of index (HTM or HEALPix) was less important than the chosen index level. "The optimum indexing level corresponds roughly to when the average search radius is of the order of the cell size."

troyraen:
I had previously pulled these changes into the `troy` branch. This PR was marked as merged and closed when I merged `troy` into `develop`. I had already deleted these changed from the `troy` branch before merging. I will open a new AllWISE PR in the future.

Issue #122, opened 2021-12-15
Connect to the Rubin alert stream integration testing deployment


troyraen:
Details and access credentials were sent to us by Eric Bellm via email. Spencer Nelson provided some additional details specific to our Kafka Connect consumer. Here are some links for reference:

- Connecting: https://github.com/lsst-dm/sample_alert_info#obtaining-the-data-with-kafka
- Schemas: https://alert-schemas-int.lsst.cloud/
- `consumer.properties` config file example: https://github.com/lsst-dm/sample_alert_info/tree/main/examples/alert_stream_integration_endpoint/java_console_consumer
- Using schema registry with Kafka Connect: https://docs.confluent.io/platform/7.0.1/schema-registry/connect.html. Spencer says, "Our stream uses Avro for the message values, not keys (we don't set the key to anything in particular), so you probably want the `value.converter` properties."
- Tools and libraries for VOEvents: https://wiki.ivoa.net/twiki/bin/view/IVOA/IvoaVOEvent#Tools_and_Libraries

PR #123, opened 2022-01-21
Connect to Rubin's test stream

troyraen:
Completes #122 

- [x] Add consumer (Kafka -> Pub/Sub connector) config files for connection to the Rubin test stream.
- [x] Document how to set up the GCP resources, run the connector, pull alerts from the resulting Pub/Sub stream and read them. This is in the README with the config files.

We can now connect to the Rubin test stream, ingest alerts, and read them out the other end. Next we should test our system under the load (#127).

### Schema management

Rubin alert packets will be Avro serialized, but the schema will not be included with the packet. There are several ways to handle this. (More details about all options are in the README.)

For now, I have:

1. Passed the alert bytes straight through the consumer (Kafka to Pub/Sub) without deserializing / converting.
2. Used `lsst.alert.packet` to load the latest Rubin schema, and `fastavro` to deserialize the alerts (in python, after pulling from the Pub/Sub stream).

In production, I think the best solution will be to:

1. Download schemas from the Confluent Schema Registry (with a `GET` request) and store them.
2. Pass the alert bytes straight through the consumer (Kafka to Pub/Sub) without deserializing / converting.
3. For each alert, check the schema version (in the Confluent Wire header), load the correct schema file and deserialize alerts using `fastavro` (in python, after pulling from the Pub/Sub stream).

The other option is to configure the Kafka Connector (i.e., our consumer) to use the Confluent Schema Registry directly. I tried to do this, but ran into the following problem. Schemas are stored in the registry under subjects and Kafka Connect is picky about how those subjects are named. Rubin has set the schema subject name to “alert-packet”, which does not conform to any of the pre-set naming strategies that Kafka Connect uses. I did not find a workaround for this issue (i.e., no way to set the subject name manually).

### Number of tasks (parallelism) in the Kafka Connector

Tasks do the actual data copying and determine the parallelism of the consumer. When I used 10 max tasks (as we do with ZTF), the consumer VM (size e2-standard-2) ran out of memory. For now I used 1 task. We should test our system under the realistic LSST load and experiment with the number of tasks, machine size etc.


troyraen:
@wmwv Requesting your review, but this is most an FYI that we can now connect to the Rubin test stream. This adds some config files and instructions, but does not alter any code that affects deployment.

broker/consumer/rubin/README.md
```diff
@@ -0,0 +1,321 @@
...
+# if successful, you will see a lot of JSON flood the terminal
+```
+
+## Run the Kafka -> Pub/Sub connector
+
+Setup:
+
+```bash
+# download the config files from broker_bucket
+sudo mkdir "$brokerdir"
```
wmwv, line 162: Brace-protect.

broker/consumer/rubin/README.md
```diff
@@ -0,0 +1,321 @@
...
+sudo gsutil -m cp -r "gs://${broker_bucket}/consumer" "$brokerdir"
```
wmwv, line 163: Brace-protect.  And I'll stop obsessively linting here.  Your usage is correct, I just often find myself using a variable, and then going back and appending something to it and always being surprised that it doesn't work, so I prefer to ${} everything.

broker/consumer/rubin/README.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+## Use the Confluent Schema Registry with the Kafka Connector
+
+Kafka Connect can use the Confluent Schema Registry directly.
+But schemas are stored under subjects and Kafka Connect is picky about how those
+subjects are named.
+See
+https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#subject-name-strategy
+**Rubin has set the schema subject name to “alert-packet”**, which does not conform
```
wmwv, line 309: What was the response from Rubin when you asked about this choice and its incompatibility making thing easy with Kafka Connect?

wmwv: Alternatively, can accept such a format be upstreamed to Kafka Connect so that it will work in the future?

docs/source/working-notes/troyraen/Rubin-stream-test/README.md
```diff
@@ -0,0 +1,309 @@
+# Connect Pitt-Google to the Rubin alert stream testing deployment
+
+THIS README HAS BEEN MOVED TO THE DIRECTORY broker/consumer/rubin.
```
wmwv, line 3: Maybe something like: the active copy of this README.md is in broker/consumer/rubin and will be kept up to date as the main LSST alert stream developments.  This README.md here is specifically for this test stream.  They are currently the same, but in the future will diverge.

wmwv (approved):
LGTM.
A couple of minor questions
Some nitpicks.

broker/consumer/rubin/README.md
```diff
@@ -0,0 +1,321 @@
...
+
+Setup:
+
+```bash
+# download the config files from broker_bucket
+sudo mkdir "$brokerdir"
+sudo gsutil -m cp -r "gs://${broker_bucket}/consumer" "$brokerdir"
+
+# set the password in two of the config files
+sudo sed -i "s/KAFKA_PASSWORD/${KAFKA_PASSWORD}/g" "${workingdir}/admin.properties"
```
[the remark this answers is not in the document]
troyraen: I believe `sudo` is necessary here because I'm operating outside of my home directory (and haven't adjusted my permissions). I `ssh`d in and manually ran these commands. If I let the machine run this as a script using its own service account it shouldn't need `sudo`.

PR #128, opened 2022-02-23
Fix bugs in metadata collection

troyraen:
Done, just waiting for testing.

- [x] Handle case where `alerts` dataframe doesn't exist for `join_metadata()`. (Set it to an empty DataFrame with correct indexes.)
- [x] Handle case where the subscription is not found. (Log the error and move on.)
- [x] Skip the upload to BigQuery if all subscriptions were empty.
- [ ] Integration test


troyraen:
@wmwv This PR contains only a few changes to a single file. These changes are already in production, so I would like to deem the "Integration test" (the only remaining to-do) as having passed, and merge this PR. So I'm requesting your review. More info follows.

### Current status

The changes here do fix the bugs they were intended to fix. The changes do not introduce adverse affects, as far as I can tell, but there is some other bug in this metadata collection process that I am currently trying to work out. It happens in/around the step that loads to BigQuery. That step fails when the script is run automatically, but I have not yet been able to reproduce the failure when I run this process manually. So for now I'm running the ZTF metadata collector manually each day while I try to find + fix the problem.

### Bigger picture

- This metadata collector has fundamental problems -- see #171. 
- Ultimately, I'd love to replace it altogether -- see #172. 
- For now, I'd like to keep it limping along. I think that the data that it's collecting is useful enough to justify this... but that's certainly debatable, so we can discuss it more if you'd like.

wmwv:
> That step fails when the script is run automatically, but I have not yet been able to reproduce the failure when I run this process manually.

Naively my first guess is a permissions problem.

wmwv (approved):
LGTM

PR #141, opened 2022-03-08
Add classifier tables

troyraen:
- [x] create tables and schemas for new tables: class_descriptions, classifiers, purity
- [x] add a data file for the class_descriptions table that maps and describes each class
- [ ] add writes to these tables from SuperNNova and extragalactic transients
- [ ] update docs
- [ ] testing


troyraen:
superceeded by #174

PR #162, opened 2022-08-25
Fix bug in data_utils.open_alert()

### LINKED ISSUE #161: Lite Dictionary Publish Message Error
**Issue Description**:
  When publishing a test message to the cloud function "ztf-lite-tag" to test the lite module, an error is encountered and the message will not be published. This error had not occurred when using broker-utils version 0.2.39. After updating the broker-utils version to 0.2.41, this error continuously occurs, not allowing a message to be published.  

**Error Message**:
" TypeError: _avro_to_dicts() got an unexpected keyword argument 'schema_map' "

Most recent call:
 File "/layers/google.python.pip/pip/lib/python3.7/site-packages/broker_utils/data_utils.py", line 117, in open_alert:
                 alert_dicts = _avro_to_dicts(alert, **kwargs)

During handling of the above exception, another exception occurred: 
" UnicodeDecodeError: 'utf-8' codec can't decode byte 0x98 in position 33: invalid start byte " 

Most recent call:
File "/opt/python3.7/lib/python3.7/json/__init__.py", line 343, in loads:
                  s = s.decode(detect_encoding(s), 'surrogatepass')

**End of Message**.

- **Broker-Utils Version**: Not found in log, but pulls from most updated version of broker-utils using line 14:
                   "from broker_utils import data_utils, gcp_utils, schema_maps, types"
   which was at the time 0.2.41
- **Trigger Topic**: ztf-alerts-tag

troyraen:
Fixes the bug referenced by #161.

Previous version of `broker_utils.data_utils.open_alert()` blindly sent `**kwargs` to the functions it called causing the wrong arguments to be sent in some cases. `open_alert()`  now sends only the arguments that are appropriate for the function it's calling.

### DIFF
--- a/broker/broker_utils/CHANGELOG.md
+++ b/broker/broker_utils/CHANGELOG.md
@@ -10,6 +10,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## \[Unreleased\]<a name="unreleased"></a>
 
+## \[0.2.42\] - 2022-08-24
+
+### Fixed
+
+- Fix bug introduced by allowing `data_utils.open_alert()` to blindly send `**kwargs` to the functions it calls. `data_utils.open_alert()` now sends only the arguments that are appropriate for the function it's calling.
+
 ## \[0.2.41\] - 2022-08-19
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


troyraen:
@hernandezc1 This fixes a bug in `open_alert()` that may affect you as well. I suggest you update your modules to use the newest version `pgb-broker-utils>=0.2.42` (change this in the requirements.txt files).

PR #170, opened 2022-09-24
Testing deployments scripts

hernandezc1:
This PR was made to test and debug the deployment scripts for the ELAsTiCC Challenge.

What we achieved:
- successfully updated IAM permissions 
- updated variables/code that were originally written for ZTF
- made corrections to code that prevented successful deployment of script


setup/setup_broker.sh
```diff
@@ -55,6 +55,7 @@ source_table="DIASource"
 if [ "$teardown" != "True" ]; then
     echo "Creating broker_bucket and uploading files..."
     gsutil mb -b on "gs://${broker_bucket}"
+    gsutil mb -b on "gs://${avro_bucket}"
```
troyraen, line 58: Make sure that you also add the appropriate command to DELETE this bucket in the code block that follows.

setup/setup_broker.sh
```diff
@@ -84,9 +85,13 @@ if [ "$teardown" != "True" ]; then
     role="userPublic"
     roleid="projects/${GOOGLE_CLOUD_PROJECT}/roles/${role}"
     user="allUsers"
+
+    # the next three lines encountered the following error: ./set_iam_policy.sh: Permission denied
```
troyraen, line 90: This is happening because I neglected to set the correct permissions on this file before I committed it. Run the following command from within the same directory (and you should lookup what the command does, if you don't know). Then commit and push.

```
chmod 744 set_iam_policy.sh
```

setup/deploy_cloud_fncs.sh
```diff
@@ -40,8 +40,10 @@ else # Deploy the Cloud Functions
     echo "Deploying Cloud Function: $store_bq_CF_name"
     store_bq_entry_point="run"
 
-    cd .. && cd cloud_functions
-    cd store_BigQuery
+    # the next two lines had the following error: No such file or directory
```
troyraen, line 43: My answer is below, **_BUT_**:
don't spend too much time on these two cloud functions. In particular, **_don't actually test either of them_**. They do not work properly yet, and they are not the focus of this PR. This PR should focus on  getting the **_Consumer_** to deploy properly (including the module's VM and pubsub stream, other resources that it relies on, the IAM permissions for its resources, etc.).

Specific answer:
My guess is that I have not `cd`'d us into the right directory here, since I changed the directory structure in the repo. So you'd just need to figure out how those calls need to be updated so that we end up in the right place.

Big picture answer:
But, I actually don't like that this script includes (because I put it there) so much `cd`ing around, for exactly this reason (it makes the code brittle). Ideally, you/we should change this script so that it just stays in one directory the whole time and points to all the paths relative to that location.

troyraen: > So you'd just need to figure out how those calls need to be updated so that we end up in the right place.

And after looking again, I see that's probably what your changes already do.

setup/setup_broker.sh
hernandezc1: Done! I got the following error, however:

Screen Shot 2022-09-26 at 7 51 48 PM

(This is when I retested the script)

troyraen: You appear (based on the name of the role) to be trying to deploy this to the GCP project "elasticc-challenge". If that is the case, try again in the correct project (the testing project).

An IAM role is itself an independent resource in google cloud. This `userPublic` role is a custom one that I created in the **_testing_** project. So no role with this name actually exists (nor should it, for now at least).

troyraen:
@hernandezc1 I am approving this now. The remaining deployment issues are fixed in #176, which builds on this. Please do the following:

- [x] Add a description of your changes here, then merge.
- [x] Then look very briefly at #175. It is mostly just syntax changes, which means lots of changes with nothing of much interest to notice.
- [x] Then look more carefully at #176. Start by reading the PRs main comment, and my commit messages (which are right below it). A lot of them are little fixes I put in to make some spot go smoother. Some are things that slipped through from here and were preventing it from quite running properly.

troyraen (approved):
See my previous comment

PR #182, opened 2022-10-22
Update Metadata Collector

### LINKED ISSUE #181: Bug: Metadata Collector - join and load
...bug in this metadata collection process that I am currently trying to work out. It happens in/around the step that loads to BigQuery. That step fails when the script is run automatically, but I have not yet been able to reproduce the failure when I run this process manually

_Originally posted by @troyraen in https://github.com/mwvgroup/Pitt-Google-Broker/issues/128#issuecomment-1257499878_

May also be in the step that joins dataframes from different streams.

troyraen:
Included changes:

- Fix #181 by dropping duplicates from the index before adding the ids to the alerts_raw metadata
- Keep duplicates when they result from publishing a message twice. This is a different type of duplicate than those causing #181. Whether we keep or drop these probably doesn't really matter -- this change is being made in order to bring the repo code into alignment with what is (and has been) in production. 
- Update resource names that were changed by #188


wmwv (approved):
LGTM

PR #183, opened 2022-10-22
Deduplicate the "alerts" stream

### LINKED ISSUE #173: Duplicate alerts
This issue documents a current problem the broker is experiencing with duplicate alerts, and starts the more general discussion of how the broker should handle duplicates.

Currently, about twice as many alerts end up in our `ztf-alerts` stream (published by our consumer) as are produced by ZTF each night. I have spent some time trying to track down where the duplicates are coming from. I think they're actually coming from ZTF, but I can't say for sure yet. 

I know that ZTF is currently having problems with duplicate alerts on their end (sometimes enough to fill up all their disk space), but they haven't been able to tell me specifically whether they're actually sending duplicates to us. I've also asked them whether they can tell if we're simply requesting/pulling the stream from them twice, but I don't have an answer on that yet either. Last night we actually received 4 times the number of alerts produced by ZTF. Half those arrived within a 45 minute window at about 7 am Pacific... very reminiscent of the stream replays that I used to see from ZTF every few days (which I "fixed" by shutting the broker down a little earlier each morning).

I've looked at various aspects of our consumer, but I haven't found any evidence that it is either requesting/pulling in the stream twice or producing duplicates internally. But I found no confirmation that it's **_not_** doing one of those things either.

The one thing I do feel confident about is that we do **_not_** have some other resource hiding somewhere in Google Cloud that is publishing to the `ztf-alerts` stream -- both because I've looked fairly thoroughly and because alerts do not end up in that stream unless our `ztf-consumer` VM is actually running.

Regardless of where they're coming from, the duplicates can be expensive for us. This is especially true when a sudden influx causes a backup in the modules that load to BigQuery, because we end up paying for stupidly-long runtimes of Cloud Functions (see https://github.com/mwvgroup/Pitt-Google-Broker/issues/96#issuecomment-929401416).

The bigger, more general question of how to make the broker handle duplicates more gracefully is also tricky. Duplicates can come from several different places, and the de-duplication approach is different for each. Worse, depending on the specifics, there may not even **_be_** a straightforward approach. To illustrate:  How would the broker even know if a given message was a duplicate? It can't check the entire data packet. It would have to track some indicator, like a Kafka offset or a Pub/Sub message ID. But if (for example) our consumer is just publishing the same message twice, those messages will have **_different_** IDs even though they contain the same data packet.

troyraen:
Fixes #173 

This PR moves the Cloud Storage module so that it sits _between_ the Consumer and the rest of the pipeline, and makes it responsible for de-duplicating the main `alerts` stream that feeds all the other modules. It off-loads the job of actually checking for duplicates to Cloud Storage itself by taking advantage of a (optional) Cloud Storage mechanism that will prevent a file from uploading if an object with the same filename already exists in the bucket.

### Specific Changes

- [x] Add a new Pub/Sub stream called `alerts_raw`. Update the Consumer to publish this stream, and the Cloud Storage module to be triggered by it.
- [x] In the Cloud Storage module:
    - [x] Add a kwarg to the file-upload API call so that it raises a `PreconditionFailed` error if an object with the same name already exists in the bucket. 
    - [x] Add a try-except block to catch the `PreconditionFailed` error. In this case, drop the duplicate alert by simply `return`ing from the Cloud Function immediately. If the error is not raised, proceed:
    - [x] Add an API call that publishes the alert to the broker's main `alerts` stream. This sends it on to be processed by the rest of the broker modules.
    - [x] Remove some log statements. (Unrelated to deduplication, but it's only a few lines and not worth a separate PR. We're logging too many things in this module and it starts to cause a quota-limit-exceeded problem at ELAsTiCC alert rates.)
- [x] Add the `alerts_raw` stream to the Metadata Collector and associated BigQuery table.

### Performance

Once I worked out the strategy, the only way to really check whether Cloud Storage is good (i.e., fast and accurate) enough at this task for this to actually work for us as a de-duplication method was to put it in production and see what happened. Turns out that Cloud Storage is very good at this. I haven't checked details yet, but the screenshots I posted in #173 show its basic performance.


broker/cloud_functions/ps_to_gcs/main.py
```diff
@@ -34,8 +36,11 @@
 
 # GCP resources used in this module
 bucket_name = f'{PROJECT_ID}-{SURVEY}-alert_avros'  # store the Avro files
+ps_topic = f"{SURVEY}-alerts"
 if TESTID != "False":
     bucket_name = f'{bucket_name}-{TESTID}'
+    ps_topic = f'{ps_topic}-{TESTID}'
```
wmwv, line 42: Could it be possible that we do want to run de-duplication under TESTEID=true?

wmwv (approved):
LGTM

One small comment to consider regarding desired behavior under test.

broker/cloud_functions/ps_to_gcs/main.py
troyraen: This line just adds the testid if there is one, but it will still de-duplicate under TESTEID=true.

PR #187, opened 2023-02-10
Ingest ZTF archive to Cloud Storage and BigQuery

troyraen:
This is a tool that downloads alerts from the [ZTF Alert Archive](https://ztf.uw.edu/alerts/public/), fixes the schemas to comply with BigQuery's strict standards, uploads the alerts to a Cloud Storage bucket, then loads the alerts from the bucket to BigQuery. For now the tool is just in my "working notes" section of the docs at: docs/source/working-notes/troyraen/ingest-ztf-archive. This PR does not touch the main pipeline.

------------ Updated 2023/04/12 ------------

The included README (at docs/source/working-notes/troyraen/ingest-ztf-archive/README.md) describes the directory contents and how to use the tool. The following is the "Instructions" subsection, copy-pasted for convenience:

Instructions

1. Setup a VM to use for the ingestion (use setup.sh)
2. Setup other GCP resources like buckets and tables if needed (use setup.sh)
3. Ingest tarballs (use ingest_tarballs.py). Basic usage:

```python
import ingest_tarballs as it

it.run(tardf=it.fetch_tarball_names(clean=["done"]))
```

This does the following:

- load list of tarballs available from [ZTF archive](https://ztf.uw.edu/alerts/public) cleaned of any that have already been logged as done
- ingest all tarballs, one at a time:
    - download and extract the tarball
    - fix the avro schemas to conform to bigquery's strict standards (parallelized)
    - upload the alerts (avro files) to the bucket (parallelized)
    - load the alerts to the table
- write logs to logs/ingest_tarballs.log and create other files in the logs directory to keep track of what's completed thru various stages

Some steps are parallelized. Defaults should work reasonably well on the SSD machine created in setup.sh. In addition, a single machine can probably do a few `it.run()` in parallel using (e.g.,) `screen` and submitting slices of `tardf` to each `run` call.

------------ Original post ------------

To run this:

- Check out this branch and then `cd` to `docs/source/working-notes/troyraen/ingest-ztf-archive`.
- Run commands in setup.sh to setup cloud resources. (Run the commands manually -- at the end you'll ssh into the VM to finish the setup)
- Start a python prompt. Here are some testing examples:

```python
from multiprocessing import Pool
from ingest_tarballs import fetch_tarball_names, ingest_one_tarball


# example that works with no errors
tarname = "ztf_public_20221216.tar.gz"  # 1.0G
ingest_one_tarball(tarname, nthreads=30)

# example that works but encounters errors (tar extraction, avro deserialization)
tarname = "ztf_public_20200127.tar.gz"  # 3.0G
ingest_one_tarball(tarname, nthreads=30)

# toy example to process tarballs in parallel.
# will need more workers and a bigger machine to make this worthwhile.
# also, since the VM I'm currently using has vcpus == threads (not cores)
# it might not even make sense to call Pool. But it does work (so far).
tardf = fetch_tarball_names()
tarsample = tardf.sample(2)
print(tarsample)
with Pool(2) as pool:
    pool.map(ingest_one_tarball, tarsample["Name"].values)
```

Here are my logs from running the above examples on a machine with 32 vcpus: [ingest_tarballs.log](https://github.com/mwvgroup/Pitt-Google-Broker/files/10706962/ingest_tarballs.log)


docs/source/working-notes/troyraen/ingest-ztf-archive/setup.sh
```diff
@@ -0,0 +1,70 @@
...
+rm Anaconda3-2022.05-Linux-x86_64.sh
+source ~/.bashrc
+conda create -n pgb python=3.7
+conda activate pgb
+# we don't use this much but it installs other dependencies we need
+pip3 install pgb-broker-utils
+pip3 uninstall fastavro
+pip3 install fastavro==1.4.4
+pip3 install lxml
+pip3 install ipython  # optional
```
wmwv, line 110: Is the longer-term plan to build an image that has Anaconda and the needed Python libraries already installed?

wmwv: My naive first thought might be to add these libraries to a requirements.txt.

docs/source/working-notes/troyraen/ingest-ztf-archive/setup.sh
```diff
@@ -0,0 +1,70 @@
...
+rm add-google-cloud-ops-agent-repo.sh
+# manually install conda, then broker_utils
+# accept conda defaults *except* say yes to last question:
+# Do you wish the installer to initialize Anaconda3 by running conda init?
+# (need conda to avoid version issues with astropy/jinja2/MarkupSafe/soft_unicode)
+wget https://repo.anaconda.com/archive/Anaconda3-2022.05-Linux-x86_64.sh
+bash Anaconda3-2022.05-Linux-x86_64.sh
+rm Anaconda3-2022.05-Linux-x86_64.sh
+source ~/.bashrc
+conda create -n pgb python=3.7
```
wmwv, line 102: Is there a reason to pin this to Python 3.7?

docs/source/working-notes/troyraen/ingest-ztf-archive/ingest_tarballs.py
```diff
@@ -0,0 +1,404 @@
...
+    logging.info(f"Alert version = {version}")
+
+    touch_schema(version, falert)
+    schema = schemas.loadavsc(version=version, nocutouts=False)
+
+    table = f"{PROJECT_ID}.{DATASET}.alerts_v{version.replace('.', '_')}"
+    touch_table(table, falert, schemas.loadavsc(version, nocutouts=True), version)
+
+    logging.info("Fixing alerts on disk")
+    with ThreadPool(nthreads) as pool:
```
wmwv: Use `Pool` instead of `ThreadPool`.  `ThreadPool` is still subject to the Global Interpreter Lock.  `Pool` is not and actually parallelize as you want.

docs/source/working-notes/troyraen/ingest-ztf-archive/ingest_tarballs.py
```diff
@@ -0,0 +1,404 @@
...
+    # Wall time: 4min 22s
+
+    # tar extraction error example
+    # %%time
+    tarname = "ztf_public_20200127.tar.gz"  # 3.0G
+    ingest_one_tarball(tarname, nthreads=30)
+
+    # possible example to process multiple tarballs in parallel
+    # since these vcpus are threads and not cores,
+    # it might not even make sense to call Pool
```
wmwv: `Pool` is what you want.  In the CPU "threads" mean a different thing than Python "threads".

Multi-threading of the CPU just creates more apparent CPUs to the Python process.

wmwv (changes requested):
1. Use `Pool` instead of `ThreadPool`
2. Figure out how to monitor the performance of the bulk upload to the Bucket.  This will be key in figuring out whether to run multiple processes on one VM or to run several VMs.

docs/source/working-notes/troyraen/ingest-ztf-archive/setup.sh
troyraen: > build an image that has Anaconda and the needed Python libraries already installed?

Possibly, but first we/I should look at what the actual requirements are at this point. A lot has changed in terms of dropping requirements that previously forced us to pin other requirements. In the meantime, there is a requirements file [here](https://github.com/mwvgroup/Pitt-Google-Broker/blob/u/tjr/update_broker_utils/broker/broker_utils/requirements_pinned.txt) where all versions are pinned that can be used to reproduce a known-to-work conda env.

Typical usage of pgb-broker-utils doesn't require the extra install/uninstall stuff in these 4 lines so I haven't worried about adding to those requirements. (But I failed to document the error that led to realizing I needed fastavro==1.4.4 and now can't remember what it was.)

docs/source/working-notes/troyraen/ingest-ztf-archive/setup.sh
troyraen: Yes. In addition to the failure chain we're avoiding here (which are astropy -> jinja2 -> MarkupSafe.soft_unicode dependency conflicts), our group has encountered issues using some of the Google API with anything other than Python 3.7 from the very beginning. But again, we should revisit this along with the other requirements issues above because our pinned-version requirements have changed.

docs/source/working-notes/troyraen/ingest-ztf-archive/ingest_tarballs.py
troyraen: Thanks, I have some ongoing confusion about exactly when/how/to what the Global Interpreter Lock is applied, so I'll read up more on that topic.

docs/source/working-notes/troyraen/ingest-ztf-archive/ingest_tarballs.py
troyraen: Thanks, I will read up more on this.

The VM that I'm creating in this PR ([n1-highcpu-32](https://cloud.google.com/compute/docs/general-purpose-machines#n1-high-cpu)) has vCPUs -- quoting from that link:

> A vCPU is implemented as a single hardware Hyper-thread on one of the available [CPU platforms](https://cloud.google.com/compute/docs/cpu-platforms).

I had assumed that meant that I could start one Python thread for each vCPU. @wmwv Do you happen to know if that is (or sounds) correct?

wmwv: Zeroth answer.  Use `Pool` to run subprocesses not `ThreadPool` to run threads.

If we replace "Python thread" by "Python subprocess" in your question:
Simple answer, sure.  That's roughly reasonable.
Practical answer:  Test it.  There's no hard limit and there's no obvious performance answer.  Try 100, try 4.  See what works.  My rough guess is that 16 will be good answer.
More theory: Processes are just things that run, they are interrupted and swapped all the time.  Your computer is running 100s of processes right now.  Almost all of your operations are I/O limited so it's not obvious how filled the CPUs are going to be.  Memory might also be the limit.  E.g., on NERSC, the rough 2 GB/core is often not enough for the things I do so I need to ask for a lot of cores just to get the memory I want.  GCP VMs are different, you specify CPU and memory separately, but the fundamental idea that you have resources of CPU, memory, network bandwidth, and dis space and different operations max out different dimensions in this space.

wmwv:
Can you add a test that exercises the reason for the `fastavro` pin?

Then if we need to update or want to figure out what's going on, we can focus on that test.

troyraen:
> Can you add a test that exercises the reason for the fastavro pin?

@wmwv It's simple but I'm not sure the best place to actually put it. Here it is:

```python
import fastavro as fa

# test alert stored with the repo
falert = "tests/test_alerts/ztf_3.3_1154308030015010004.avro"

# standard fastavro call to deserialize into a list of dicts
with open(falert, "rb") as fin:
    list_of_dicts = list(fa.reader(fin))
```

Wtih `fastavro==1.4.4`, this works. With `fastavro==1.7.3`, this fails with `TypeError: float() argument must be a string or a number, not 'NoneType'`. 

This is very likely a ZTF-specific problem, due to the non-standard ordering of their Avro schema types (the same problem that causes us to do a "fix schema" step before storing the Avros in the bucket).

Correcting one thing I said before: There is one module in the main pipeline that requires this to be pinned and that's the cloud function that uploads to the bucket. That's the only module that has to deal with the ZTF schema. It's already pinned in that requirements.txt file. I added a note there that better documents the reason it's pinned and how to reproduce the error.

troyraen:
@wmwv I think this is ready for final review. The archive ingestion job is done, except that some alerts have not yet been loaded from the bucket to the table due to the bug in BigQuery. My understanding is that the bug-fix is out and we're just waiting for it to hit our region. When it does, I'll have to finish the load jobs.

wmwv (changes requested):
Thank you for documenting the pin for `fastavro`.  I encourage you to copy the lines of the test in the comment into a test in `tests/` to allow us to easily monitor if the fastavro problem resolves at some point and to easily catch if someone later tries to remove the pin and shouldn't.

wmwv (approved):
But this is an optional request.  You can merge as is for now and start a new issue to add the test.

PR #188, opened 2023-04-06
Change resource names

troyraen:
These are changes to the pipeline that were prompted by #187. Buckets and tables storing alerts are now separated by Avro schema version.

- Change BigQuery dataset name: "ztf_alerts" -> "ztf"
- Change BigQuery table name storing alerts: "alerts" -> "alerts_v3_3"
- Change bucket name storing alerts: "{project_id}-ztf-alert_avros" -> "{project_id}-ztf_alerts_v3_3"
- Change the region for datasets and buckets: multi-region -> single region (default us-central1)
- Set IAM permissions on dataset and bucket granting public access. Make the bucket requestor-pays.

(Example names are given for survey = "ztf" and Avro schema version "3.3")


broker/setup_broker/deploy_cloud_fncs.sh
```diff
@@ -9,6 +9,7 @@ teardown="${2:-False}"
 # "True" tearsdown/deletes resources, else setup
 survey="${3:-ztf}"
 # name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
```
wmwv, line 12: Can you explain the "v3_3"?  Where does that come from?  Would that need to be changed in some future version?

broker/setup_broker/setup_gcp.py
```diff
@@ -93,7 +93,7 @@
 PROJECT_ID = os.getenv('GOOGLE_CLOUD_PROJECT')
 
 
-def _resources(service, survey='ztf', testid='test'):
+def _resources(service, survey='ztf', testid='test', versiontag="v3_3"):
```
wmwv, line 96: Is there a easier place to define this "v3_3" string.  When this has to be updated, it will be difficult to find on line 96 of a file.

broker/setup_broker/setup_gcp.py
```diff
@@ -224,7 +225,7 @@ def _confirm_options(survey, testid, teardown):
         sys.exit(msg)
 
 
-def setup_bigquery(survey='ztf', testid='test', teardown=False) -> None:
+def setup_bigquery(survey='ztf', testid='test', teardown=False, versiontag="v3_3", region="us-central1") -> None:
```
wmwv, line 236: Same versiontag default definition comment.

broker/setup_broker/setup_gcp.py
```diff
@@ -382,7 +386,7 @@ def _setup_dashboard_resource_names(survey='ztf', testid='test'):
     return resource_maps
 
 
-def setup_buckets(survey='ztf', testid='test', teardown=False) -> None:
+def setup_buckets(survey='ztf', testid='test', teardown=False, versiontag="v3_3", region="us-central1") -> None:
```
wmwv, line 407: versiontag default

wmwv (changes requested):
Looks good.  

One issue about making the versiontag "v3_3" string more prominently defined and commented with a note about when it should be updated.

And then I agree with Codacy's helpful reminders about shell scripts.  I have to admit that I find linters in Python _useful_, but I find linters in shell script _transformative_ -- my own shell scripting habits could use a lot of work.

broker/setup_broker/setup_broker.sh
```diff
@@ -10,7 +10,7 @@ survey="${3:-ztf}"
 # name of the survey this broker instance will ingest
 # 'ztf' or 'decat'
 schema_version="${4:-3.3}"
-versiontag="v$(echo ${schema_version} | tr . _)"  # 3.3 -> v3_3
+versiontag=v$(echo "${schema_version}" | tr . _)  # 3.3 -> v3_3
```
wmwv, line 13: And while it's definitely good to quote all these string expansions, is there some point along the way that expects (requires?) that the versiontag not have any whitespace?

Not a question to fix, just a question to consider in your own mind for thinking about shell scripting and documenting behavior for the future.

wmwv, line 13: And while it's definitely good to quote all these string expansions, is there some point along the way that expects (requires?) that the versiontag not have any whitespace?

Not a question to fix, just a question to consider in your own mind for thinking about shell scripting and documenting behavior for the future.

wmwv (approved):
LGTM

broker/setup_broker/deploy_cloud_fncs.sh
troyraen: Thanks for pointing that out, I've updated the docs.

broker/setup_broker/setup_gcp.py
troyraen: I don't really understand what you're asking. In practice, this gets defined by the user and passed into the script. Are you wanting to more easily update the default value? I think that's a problem throughout this script... I agree that the way this entire script is setup is bad, but I don't see a reason to make a special (i.e., different than `survey` or `testid`) way to set the `versiontag` default. (I'm guessing that this script will be entirely revamped or replaced in the near future anyway, along with the rest of the setup scripts.)

broker/setup_broker/setup_broker.sh
troyraen: (Yes, GCP will require no whitespace in `versiontag`, since it becomes part of resource names.)

broker/setup_broker/setup_gcp.py
wmwv: For this PR, don't change anything. 

But for design consideration in the future:

For things like this I might consider having some globals (or at least module-level variables) for defaults.  I tend to use all caps for these to indicate that they are special values.

VERSIONTAG = v3_3

They can live up at the top of the file, ideally only in one file.  It's a bit annoying to share between shell scripts and Python code, but you could at least reduce to two files.

One reason I call out versiontag in particular is that survey, testid, and versiontag each draw from a different part of the understanding.  Survey makes immediate sense, "Oh, right I need to specify the survey to listen to/load."  Testid makes a little sense if you think you're testing.  But a user has no particularly obvious way to reason about what versiontag should be.

One might also decide to not provide defaults and force to user to immediately have to make choices.

wmwv (approved):

PR #189, opened 2023-04-11
Managing Microservices w/ Dev Environment

hernandezc1:
The motivation for this PR is to configure a dev stage where each microservice (i.e., cloud function) is independently deployed by its own deployment script. 

The following changes have been made:
- [x] each cloud function has its own deployment script named `deploy.sh`
- [x] each cloud function has the following directory structure: `/broker/cloud_functions/cloud_function_name/deploy.sh`
- [x] `deploy_cloud_fncs.sh` has been deleted
- [x] `setup_broker.sh` has been updated to call each of the individual deployment scripts
- [x] the `schema_maps` directory and its files have been included in the directories of the following cloud functions: `ps_to_gcs` and `store_BigQuery`. These resources are needed to make the cloud function(s) deployable. 

The changes have been tested. Each cloud function was successfully deployed.


broker/cloud_functions/ps_to_gcs/decat.yaml
troyraen, line 1: Looking at the directory structures, I think I'd prefer to have the schema maps in a dedicated "schema_maps" subdirectory. I think that would make it more self-explanatory about what those files are for, in addition to reducing clutter as more are added.

broker/cloud_functions/check_cue_response/deployment.sh
```diff
@@ -0,0 +1,41 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+zone="${CE_ZONE:-us-central1-a}" # use env variable CE_ZONE if it exists
```
troyraen:
```suggestion
zone="${5:-us-central1-a}"
```

As of #192, we no longer use the `CE_ZONE` environment variable, so let's change it to `5` so that the script accepts it as an input argument. If the variable isn't actually being used in the script (as Codacy indicates here, and in other files below) it should just be removed altogether.

broker/cloud_functions/classify_snn/deployment.sh
```diff
@@ -0,0 +1,43 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+zone="${CE_ZONE:-us-central1-a}" # use env variable CE_ZONE if it exists
```
troyraen:
```suggestion
zone="${5:-us-central1-a}"
```

or delete if unused.

broker/cloud_functions/lite/deployment.sh
```diff
@@ -0,0 +1,41 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+zone="${CE_ZONE:-us-central1-a}" # use env variable CE_ZONE if it exists
```
troyraen:
```suggestion
zone="${5:-us-central1-a}"
```

or delete if unused.

broker/cloud_functions/ps_to_gcs/deployment.sh
```diff
@@ -0,0 +1,43 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+zone="${CE_ZONE:-us-central1-a}" # use env variable CE_ZONE if it exists
```
troyraen:
```suggestion
zone="${5:-us-central1-a}"
```

or delete if unused.

broker/cloud_functions/store_BigQuery/deployment.sh
```diff
@@ -0,0 +1,41 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+zone="${CE_ZONE:-us-central1-a}" # use env variable CE_ZONE if it exists
```
troyraen:
```suggestion
zone="${5:-us-central1-a}"
```

or delete if unused.

broker/cloud_functions/tag/deployment.sh
```diff
@@ -0,0 +1,43 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+zone="${CE_ZONE:-us-central1-a}" # use env variable CE_ZONE if it exists
```
troyraen:
```suggestion
zone="${5:-us-central1-a}"
```

or delete if unused.

broker/setup_broker/setup_broker.sh
```diff
@@ -129,4 +129,28 @@ fi
 #--- Deploy Cloud Functions
 echo
 echo "Configuring Cloud Functions..."
-./deploy_cloud_fncs.sh "${testid}" "${teardown}" "${survey}" "${versiontag}" "${zone}"
+cd .. && cd cloud_functions
```
troyraen: I started some bad trends in these shell scripts and `cd`ing around a lot is one of them. Unfortunately it's not easy to fix here because the cloud function gets deployed from the current working directory, so we need to `cd` into it. We could avoid the `cd` by using the `--source` flag in the [`gcloud functions deploy` command](https://cloud.google.com/sdk/gcloud/reference/functions/deploy#--source), but it would have to be set in each `deployment.sh` script and be relative to whatever the current working directory is -- that is generally `broker/setup_broker` before we start `cd`ing around, but that seems like a weird, brittle, relative path to set in those files.

Codacy is complaining about this, so for now I think we should fix it by just accepting Codacy's suggestion and doing 

```suggestion
cd .. && cd cloud_functions || exit
```

and doing similar in all the `cd`s that follow.

broker/cloud_functions/check_cue_response/deployment.sh
troyraen, line 1: Not hugely important, but I would prefer to call these scripts "deploy.sh". This is because I prefer to use a verb to name scripts like this.

broker/cloud_functions/ps_to_gcs/deployment.sh
troyraen, line 1: When using this script as a stand-alone, the cloud function does not deploy successfully because the bucket that holds the Avro files does not get created. Currently the bucket is created in `setup_gcp.py` and then permissions are set in `setup_broker.sh`.

There's a few ways to fix this. One would be to just add the relevant commands to this script. (The needed cli commands are [`gsutil mb`](https://cloud.google.com/storage/docs/gsutil/commands/mb) plus the permissions commands in `setup_broker.sh`.) Another would be to create a separate script that creates the bucket and sets permissions, and call it from this script and also from the main `setup_broker.sh` script.

troyraen (changes requested):
Nice work. A few changes to be made still. 

1. Fix the issues Codacy is identifying
2. The ps_to_gcs cloud function doesn't currently deploy when the development.sh script is used alone. See comment below.
3. We no longer use the environment variable `CE_ZONE`. See comment below.
4. Looking at the directory structures, I think I'd prefer to have the schema maps in a dedicated "schema_maps" subdirectory. See comment below.

broker/cloud_functions/check_cue_response/deployment.sh
hernandezc1: I reviewed the bash scripts (`deploy.sh`) and python scripts (`main.py`) for each of the individual cloud functions. It seems that `check_cue_response` is the only cloud function which uses `zone`. Previously, `zone` is used in `main.py`, but not `deploy.sh`. I've since added `zone` as an environment variable that gets passed when the command `gcloud functions deploy "${check_cue_CF_name}" \` is run in `deploy.sh`

broker/cloud_functions/check_cue_response/deploy.sh
```diff
@@ -37,5 +37,5 @@ else # Deploy the Cloud Functions
         --entry-point "${check_cue_entry_point}" \
         --runtime python37 \
         --trigger-topic "${check_cue_trigger_topic}" \
-        --set-env-vars TESTID="${testid}",SURVEY="${survey}",VERSIONTAG="${versiontag}"
+        --set-env-vars TESTID="${testid}",SURVEY="${survey}",VERSIONTAG="${versiontag}",ZONE="${zone}"
```
hernandezc1, line 41: this is what I was referring to in the previous comment: https://github.com/mwvgroup/Pitt-Google-Broker/pull/189#discussion_r1195654338

broker/cloud_functions/ps_to_gcs/deployment.sh
hernandezc1: The cloud function now deploys successfully. I added the relevant `gsutil mb` commands to this script and included the permission commands from `setup_broker.sh`. `deploy.sh` checks to see if the necessary buckets exist, and creates them if they do not

troyraen:
I'm confused about why the "broker_bucket" stuff is being moved? I think that should not be something that the ps_to_gcs cloud function controls. That bucket is used much more broadly.

broker/cloud_functions/check_cue_response/deploy.sh
troyraen: Nice. From what I can tell, this was a bug in the script.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,71 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+PROJECT_ID="${5:-avid-heading-329016}"
```
troyraen: I think this should be an environment variable rather than a script argument.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,71 @@
...
+
+#--- GCP resources used in this script
+avro_bucket="${PROJECT_ID}-${survey}_alerts_${versiontag}"
+avro_topic="projects/${PROJECT_ID}/topics/${survey}-alert_avros"
+broker_bucket="${PROJECT_ID}-${survey}-broker_files"
```
troyraen: I don't think this module needs to touch this bucket at all.

broker/cloud_functions/ps_to_gcs/schema_maps/README.md
```diff
@@ -0,0 +1,7 @@
+# Schema Maps
+
+The files in this directory contain mappings between the schema of an individual survey and a PGB-standardized schema that is used within the broker code.
+
+Note: This directory is __not__ packaged with the `broker_utils` module.
+In order to allow broker instances to use unique schema maps, independent of other instances, each instance uploads this directory to its [`broker_files`] Cloud Storage bucket upon setup.
+The broker code loads the schema maps from the bucket of the appropriate instance at runtime.
```
troyraen: This paragraph doesn't make sense here.

broker/cloud_functions/store_BigQuery/schema_maps/README.md
```diff
@@ -0,0 +1,7 @@
+# Schema Maps
+
+The files in this directory contain mappings between the schema of an individual survey and a PGB-standardized schema that is used within the broker code.
+
+Note: This directory is __not__ packaged with the `broker_utils` module.
+In order to allow broker instances to use unique schema maps, independent of other instances, each instance uploads this directory to its [`broker_files`] Cloud Storage bucket upon setup.
+The broker code loads the schema maps from the bucket of the appropriate instance at runtime.
```
troyraen: This paragraph doesn't make sense here.

troyraen (changes requested):
- Fix the Codacy issues
- There is a warning: "This branch is out-of-date with the base branch". Follow the instructions there to fix it.
- There are other resources that each module depends on that are not being created in the deploy scripts, like Pub/Sub topics and BigQuery tables. What is the plan there? Are the deploy scripts intended to make the modules functional? Or only deployable, and then the setup_gcp.py script also needs to be run for the modules to be functional?
- The broker_files bucket is still being created by the ps_to_gcs deployment script. I don't think that module needs that bucket at all. Since the schema maps have been moved into each cloud fnc directory that needs them, I think the only modules that need that bucket are the consumer and the night conductor.

I will test the changes once we sort out the above.

broker/cloud_functions/ps_to_gcs/deploy.sh
hernandezc1: I've removed any commands that call on broker_bucket

hernandezc1:
At our CI/CD meeting last week, we discussed whether we want the deploy.sh scripts to make the modules functional or deployable. We decided that since `setup_broker.sh` creates the resources that some of the modules depends on, the `deploy.sh` scripts should only check to see if those resources exist, and notify the user if they do not.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,51 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-ztf}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-v3_3}"
+PROJECT_ID="${PROJECT_ID:-avid-heading-329016}"
```
troyraen:
```suggestion
PROJECT_ID=$GOOGLE_CLOUD_PROJECT # get the environment variable
```

This needs to come from the environment variable `GOOGLE_CLOUD_PROJECT`, and it shouldn't get a default value.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,51 @@
...
+    if [ "${testid}" != "False" ]; then
+        gcloud functions delete "${ps_to_gcs_CF_name}"
+    fi
+
+else # Deploy the Cloud Functions
+
+    #--- Check to see if required resources exist
+    if ! gsutil ls -b "gs://${avro_bucket}/" >/dev/null 2>&1; then
+        echo "gs://${avro_bucket} does not exist. Deploy setup_broker.sh to create the required resources"
+    fi
```
troyraen: I think the idea for this PR is that these deploy.sh scripts should run successfully when used on their own. The modules will not be _functional_ because the other resources won't exist (like the trigger topic/subscription), but the Cloud Functions should _deploy_ successfully.

This `avro_bucket` is required to exist in order for this Cloud Function to deploy successfully. So this script needs to actually create this bucket. I think this is what's needed:

```suggestion
    # create the bucket that will store the alerts
    gsutil mb -l "${region}" "gs://${avro_bucket}"
    gsutil uniformbucketlevelaccess set on "gs://${avro_bucket}"
    gsutil requesterpays set on "gs://${avro_bucket}"
    gcloud storage buckets add-iam-policy-binding "gs://${avro_bucket}" \
        --member="allUsers" \
        --role="roles/storage.objectViewer"
    format=json  # json or none; if json, file metadata sent in message body
    gsutil notification create \
        -t "$avro_topic" \
        -e OBJECT_FINALIZE \
        -f "$format" \
        "gs://${avro_bucket}"
```

The `region` will need to be passed in as a script argument. The `avro_topic` will need to be defined above with the other "GCP resources used in this script". The bucket also needs to be deleted in the teardown section above.

Since this the only module that needs this bucket, let's remove the bucket setup from the scripts in the setup_broker directory. So the above lines should be removed from setup_broker.sh, and the bucket should be removed from setup_gcp.py (line 148).

troyraen (changes requested):
There's a couple issues with the ps_to_gcs cloud function. Everything else looks good and runs as expected.

broker/cloud_functions/ps_to_gcs/deploy.sh
troyraen, line 1: I just noticed that running this script actually creates the two topics. Make it clean up after itself (delete them). Test the other scripts for the same issue and fix them as well. Important not to leave orphaned resources laying around the project.

troyraen (changes requested):
One issue I saw with deleting resources. Everything else seems to be working.

broker/cloud_functions/ps_to_gcs/deploy.sh
hernandezc1: I've added the `gcloud` commands to `deploy.sh` to delete the two topics. Since the topics are already deleted when setting `teardown="True"` and running the `setup_broker.sh` script, I'm getting the following error message(s)
`ERROR: Failed to delete topic [projects/avid-heading-329016/topics/ztf-alert_avros-(enter testid)]: Resource not found (resource=ztf-alert_avros-(enter testid)).
ERROR: (gcloud.pubsub.topics.delete) Failed to delete the following: [ztf-alert_avros-(enter testid)].`

This is technically fine, just wanted to point it out. I ran an additional test where instead of setting `teardown="True"` and running `setup_broker.sh`, I ran `deploy.sh` for `ps_to_gcs`. All the relevant resources were deleted and no error messages popped up (as expected).

hernandezc1: After running some tests, I believe there are no additional orphaned resources (and everything that needs to be deleted gets deleted).

troyraen (approved):
LGTM

PR #192, opened 2023-04-27
Assign static IP to Consumer VM

troyraen:
This is being done in response to a request from ZTF for an IP address that they can whitelist, which will be necessary for an upcoming changes on their end.

Primary change:

- Assign a static IP address to the Consumer VM.

Other changes:

- Remove the `CE_ZONE` environment variable. This "zone" is closely related to the "region" that is a setup-script argument, and it's hard to enforce consistency between the two when they come from two different places.
- Two bug-fixes in the consumer's shell scripts

@wmwv FYI
@djperrefort will you review?


djperrefort (approved):
LGTM

PR #195, opened 2023-05-22
Adding ELAsTiCC schema(s) version 0.9.1 to broker_utils

hernandezc1:
ELAsTiCC has updated their schema to [version 0.9.1](https://github.com/LSSTDESC/elasticc/tree/main/alert_schema) for the start of their test stream (May 2023). Note, this is not the start of ELAsTiCC2 and is merely a test. 

This PR focuses on making updates to `broker_utils` to accommodate for the new schema version. Updates include:
- [x] adding the new schema files to the `broker_utils/avro_schemas` directory
- [x] registering the new schemas in `load.py`
- [x] updating the README.md in the `broker_utils/avro_schemas` directory
- [x] updating the CHANGELOG.md file

Note: The changes made in this PR have been tested to see whether we'd be able to open an alert from the `elasticc-loop` topic using the updated schema. The test was successful


troyraen (changes requested):
I think this is good but I would like to actually test it once we have an incoming alert with the new schema. One more thing that needs to be done for any change to broker_utils: update the changelog.

We will also need to bump the version and publish to PyPI. Let's talk about that after we can actually test this with an alert.

hernandezc1:
The changes you requested have been completed. We still need to run the tests you mentioned in the previous comment

troyraen (approved):
Looks good

troyraen:
I published to PyPI, so these changes are now available in pgb-broker-utils version 0.2.47. @hernandezc1 merge this, then take a look at #198.

PR #196, opened 2023-05-24
Update main/elasticc to reflect microservice structure

hernandezc1:
Following the work done in PR #189, this PR makes the following changes:

- [x] previously, `store_BigQuery` commands had been commented out. These commands have been reinstated
- [x] `ps_to_gcs` and `store_BigQuery` have the following directory structure: `/broker/cloud_functions/cloud_function_name/deploy.sh`
- [x] each cloud function has its own deployment script named `deploy.sh`
- [x] `deploy_cloud_fncs.sh` has been deleted
- [x] `main.py` for `ps_to_gcs` has been updated to accommodate version 0.2.47 of `broker_utils`
- [x] `requirements.txt` for `ps_to_gcs` and `store_BigQuery` have been updated to accommodate version 0.2.47 of `broker_utils`
- [x] `setup_broker.sh` has been updated to call each of the individual deployment scripts
- [x] the `schema_maps` directory and its files have been included in the directories of the following cloud functions: `ps_to_gcs` and `store_BigQuery`. These resources are needed to make the cloud function(s) deployable. 

The changes have been tested. Each cloud function was successfully deployed and all the resources are deleted when `teardown="True"`.


troyraen:
> This PR serves two purposes:
> 
> * to make any updates to the `main/elasticc` repo as a result of the new schemas
> * and to slowly prepare `main/elasticc` to be merged with the `develop` branch

I recommend that these be two separate PRs. Updating for the new schemas is easily separable from the other stuff and probably needs to be done more quickly.

hernandezc1:
Currently, I've removed the store_BigQuery resources because they are not being used. I'm not sure if these resources will be used in the future, however (i.e., is it worth keeping in the repo?).

hernandezc1:
Hi @troyraen, I assigned you as a reviewer for this PR and I think it simultaneously requested a review. This PR still isn't ready to be reviewed

troyraen (changes requested):
I've just looked at the differences between the branches `develop` and `main/elasticc`... The differences are major and we're going to have to think more carefully about how to merge them.

If you want to make these changes to the `main/elasticc` branch because you actually want to use them for elasticc, then we can make that work. But in that case let's finish #189 first and then bring this in line with it and go from there.

However, if the only purpose here is to prepare to merge `main/elasticc` into `develop`, unfortunately I think this is only going to add to the confusion. In that case, I would prefer to close this PR and plan a more wholistic approach.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,77 @@
...
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-elasticc}"
+# name of the survey this broker instance will ingest
+max_instances="${4:-500}"
+BROKER_VERSION="${5:-0.1.0e}"
```
troyraen: I think it does not make sense for this to get a default value. Consider just pulling in from the version file right here rather than sending it in as an argument.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,77 @@
...
+# This script will not delete Cloud Functions that are in production
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-elasticc}"
+# name of the survey this broker instance will ingest
+max_instances="${4:-500}"
```
troyraen: I think I'm in favor of removing this option. We added it as a safety mechanism for our very first ELAsTiCC tests, but I don't think it's really necessary to keep around. Check the metrics for the average and max number of instances used during normal ingestion and let me know what you think.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,77 @@
...
+    avro_topic="${avro_topic}-${testid}"
+    ps_to_gcs_trigger_topic="${ps_to_gcs_trigger_topic}-${testid}"
+    ps_to_gcs_CF_name="${ps_to_gcs_CF_name}-${testid}"
+fi
+
+if [ "${teardown}" = "True" ]; then
+    # ensure that we do not teardown production resources
+    if [ "${testid}" != "False" ]; then
+        gsutil rm -r "gs://${avro_bucket}"
+        gcloud pubsub subscriptions delete "${avro_topic}-reservoir"
```
troyraen, line 33: This script should not touch this subscription.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,77 @@
...
+    roleid="projects/${GOOGLE_CLOUD_PROJECT}/roles/userPublic"
+    trigger_event=OBJECT_FINALIZE
+    format=json  # json or none; if json, file metadata sent in message body
+    gsutil notification create \
+        -t "$avro_topic" \
+        -e "$trigger_event" \
+        -f "$format" \
+        "gs://${avro_bucket}"
+    gcloud pubsub topics add-iam-policy-binding "${avro_topic}" --member="${user}" --role="${roleid}"
+    gcloud pubsub subscriptions create "${avro_topic}-reservoir" --topic "${avro_topic}"
```
troyraen, line 62: This script should not touch this subscription.

broker/cloud_functions/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,77 @@
...
+    ps_to_gcs_CF_name="${ps_to_gcs_CF_name}-${testid}"
+fi
+
+if [ "${teardown}" = "True" ]; then
+    # ensure that we do not teardown production resources
+    if [ "${testid}" != "False" ]; then
+        gsutil rm -r "gs://${avro_bucket}"
+        gcloud pubsub subscriptions delete "${avro_topic}-reservoir"
+        gcloud pubsub topics delete "${avro_topic}"
+        gcloud pubsub topics delete "${ps_to_gcs_trigger_topic}"
```
troyraen, line 35: Does this script actually create this topic? Other modules depend on it so I'm leery of deleting it here. (I know we ran into some topic(s) getting automatically created but I don't remember if it was this one.)

broker/cloud_functions/ps_to_gcs/requirements.txt
```diff
@@ -3,7 +3,29 @@
 # dependencies for a Cloud Function must be specified in a `requirements.txt`
 # file (or packaged with the function) in the same directory as `main.py`
 
-fastavro
+fastavro==1.4.4  # fastavro==1.4.4 is known to work for ztf (see below)
```
troyraen, line 6: Is this really necessary for ELAsTiCC? We want to avoid pinning to old versions like this if at all possible.

broker/cloud_functions/store_BigQuery/README.md
```diff
@@ -1,3 +1,3 @@
 # Store alerts in BigQuery
 
-This Cloud Function stores alert data in multiple BigQuery tables.
+This Cloud Function stores alert data in multiple BigQuery tables.
```
troyraen, line 3: This is not a real change. Try checking out the file again to cleanly restore the original version:

```
git checkout main/elasticc -- broker/cloud_functions/store_BigQuery/README.md
```

broker/cloud_functions/store_BigQuery/main.py
```diff
@@ -174,4 +174,4 @@ def publish_pubsub(alert_dict: dict, table_dicts: Dict[str, Optional[dict]]):
     msg = b""
 
     # publish
-    gcp_utils.publish_pubsub(ps_topic, msg, attrs=attrs)
+    gcp_utils.publish_pubsub(ps_topic, msg, attrs=attrs)
```
troyraen, line 177: Same as above. This is not a real change.

setup/setup_broker.sh
```diff
@@ -5,6 +5,7 @@ teardown="${2:-False}"  # "True" tearsdown/deletes resources, else setup
 survey="${3:-elasticc}"  # name of the survey this broker instance will ingest
 max_instances="${4:-500}"  # max N of concurrent Cloud Fnc instances (per deployed module)
```
troyraen: As above, I'm in favor of removing this.

setup/setup_broker.sh
```diff
@@ -72,47 +65,44 @@ if [ "${teardown}" != "True" ]; then
     # create dashboard
     gcloud monitoring dashboards create --config-from-file="templates/dashboard.json"
     # create bigquery
-    # bq mk --dataset "${bq_dataset}"
-    # bq mk --table "${bq_dataset}.${alerts_table}" "templates/bq_${survey}_${alerts_table}_schema.json"
-    # bq mk --table "${bq_dataset}.${source_table}" "templates/bq_${survey}_${source_table}_schema.json"
+    alerts_table="alerts"
+    source_table="DIASource"
```
troyraen: These needed to be defined in the section above called "GCP resources used directly in this script"

troyraen (changes requested):
Some things I noticed from reading through the changes. I have not tried to run anything yet.

broker/cloud_functions/ps_to_gcs/deploy.sh
troyraen: See #201

broker/cloud_functions/ps_to_gcs/requirements.txt
hernandezc1: I don't believe it is, but it is defined that way in the `develop` branch. I was trying to keep things consistent between branches

troyraen: Ok, let's leave it unpinned here and deal with the conflict if/when we merge the two branches.

PR #202, opened 2023-07-13
Add SuperNNova classification to Pub/Sub metadata

troyraen:
In the spirit of making it easy to filter the stream, classifiers should add their classification to the Pub/Sub message metadata. The SuperNNova module was not doing this, so this PR makes that change.

### DIFF
--- a/broker/cloud_functions/classify_snn/main.py
+++ b/broker/cloud_functions/classify_snn/main.py
@@ -64,7 +64,11 @@ def run(msg: dict, context) -> None:
     snn_dict = _classify_with_snn(alert_lite)
 
     # announce to pubsub
-    gcp_utils.publish_pubsub(ps_topic, {**alert_lite, "SuperNNova": snn_dict}, attrs=attrs)
+    gcp_utils.publish_pubsub(
+        ps_topic,
+        message={**alert_lite, "SuperNNova": snn_dict},
+        attrs={**attrs, "supernnova_class": str(snn_dict["predicted_class"])},
+    )
 
     # store in bigquery
     errors = gcp_utils.insert_rows_bigquery(


wmwv:
1. Are there cost or performance implications of this change?
2. How will this work in the overall diagram if someone wants to listen to just the SuperNNova-classified stream, or wants to wait for a stream that has a specific set of classifiers run?  Where does this logic live to create a Topic that has a defined number of classifications -- are the reasonable choices 0 and "all"?

troyraen:
> Are there cost or performance implications of this change?

No.

I don't understand the next questions...

> How will this work in the overall diagram if someone wants to listen to just the SuperNNova-classified stream, or wants to wait for a stream that has a specific set of classifiers run? 

If someone wants just the SuperNNova classifications then they can just listen to the stream that module produces. If/when we add more classifiers, assuming they are lined up serially in the pipeline, a user could choose to listen to the stream produced by the last classifier in the pipeline that they actually want. All previous modules will have attached their results to the message (class prediction in the metadata, details in the message packet) before passing it along to the next module. Does that answer the question?

> Where does this logic live to create a Topic that has a defined number of classifications -- are the reasonable choices 0 and "all"?

I really don't understand this question. Can you rephrase?

wmwv:
Why should the classifiers be lined up serially is really my question.

I think we have a difference in design idea here.  There's no particular dependency between classifiers so there's no particular logical reason to order them in any particular way.  Thus, having a series of topics to listen to with implicitly different amounts of information is a difficult thing for a user to figure out.

However, if the classifications are parallel, though, then it becomes more difficult to say: "I want to listen to a stream where these 5 classifiers have been run." because there's no explicitly thing that guarantees or could construct such a thing.

troyraen:
Yes, I see and totally agree with your summary of the problem. I think I've resigned to running them serially exactly because I haven't been able to figure out a good way to join the results of classifiers run in parallel. 

I suppose (but don't think I've actually looked) that Google would recommend Dataflow. I'm skeptical (obviously), but we could always try it. 

Perhaps another option would be to run a module that periodically queries BigQuery and produces the joined stream that way. It could listen to the main "alerts" stream, wait 5(?) seconds, then pull whatever classifications have made it to BigQuery by then, add them to the message, then publish. This sounds expensive both because we'd pay significant compute costs for the waiting, then also the cost of so many small queries. So, perhaps this should be running on a VM, and not listening to an input stream but rather just doing a periodic query for everything loaded since the last query and producing a stream from that. We'd have to really think through the details of exactly what we'd query for and how we structure our BigQuery tables, but I like this solution the best so far.

jrossthomson:
I'm interested to understand this. Generically, each ztf/ruben message
will have associated attributes, some of which are intrinsic, some of which
are generated, for example by a classifier. If there was a single
processing stream where messages and attributes were published after each
processing step, it would be possible to select for processing from this
stream only messages having the desired attributes.

Effectively, it would manage some kind of graph, where each processing node
would run only on the messages with the required attributes set. This would
allow dependencies to be managed. We might want to think about how much
this would cost.

Is the issue that non-Pitt/Google processors would not have publishing
rights to this stream?

--



*J Ross Thomson, PhD*

***@***.***

HPC & Scientific Computing

‪412-345-6667‬

<https://orcid.org/0000-0002-9593-7052>

troyraen:
@jrossthomson That's interesting, if I'm understanding it correctly. So the idea is that we would produce just one stream for a given survey and every module would publish to it, instead of every module publishing to separate streams?

It would take some thought to design. For example, what attributes would the modules be required to add and what does the filter actually look like that picks out the right messages? Looking at the [filter syntax](https://cloud.google.com/pubsub/docs/subscription-message-filter#filtering_syntax) I'm coming up with some ideas, but would need to take more time to think it all the way through. But it's doable.

> This would allow dependencies to be managed.

That would help a lot, especially in deployment. I like the simplification of managing one stream instead of Order(10) streams. 

> Is the issue that non-Pitt/Google processors would not have publishing rights to this stream?

No, I think that is not a problem.

> We might want to think about how much this would cost.

That is the biggest potential problem I see. We will incur throughput charges for every message in the pipeline to be delivered to every module in the pipeline, regardless of whether that module actually selects and processes a given message. I think we're going to be in the regime where that cost is significant. But we could check.

troyraen:
> > Are there cost or performance implications of this change?
> 
> No.

To be more specific, yes, but I think they're negligible.

troyraen:
Just talked to @Crash-GHaun who suggested that the modules could also share a subscription. He is familiar with others doing it this way. I'm guessing the cost would be cheaper overall (vs different subscriptions for every module), but estimating is much less straightforward here. More details:

The filter would have to be in the function itself, rather than a filter natively on the subscription. So, every message in the shared subscription would trigger every function. Then functions would check the message attributes and immediately send a nack (not acknowledge) and return if they should not process a given message. Else, continue processing and send an ack at the end.

One complication is that two modules cannot share a subscription if they both need to process the same messages. So, anything that runs parallel to another module(s) in the pipeline needs a separate subscription.

jrossthomson:
> 

I think this sounds great. It might be right to have a single "raw" stream, where you are not expecting to filter, and a "processed" stream, with all the downstream products. 

For third party processors, a registration process to get blessed attributes into the stream, and other requirements.

One more question is, do the intermediate products get stored long term? Or is the short term storage on Pubsub the limit? Perhaps a bucket with a time-to-live for intermediate products and only messages with a specific signature get stored long term.  Third party processors could store whatever they choose, of course.

troyraen:
@Crash-GHaun suggested using [Redis](https://redis.io/) to cache messages (in addition to publishing to Pub/Sub). This could be used to join the results of classifiers running in parallel. Each module could check the cache for previously published results for the given alert and add the info to its own message before publishing.

I keep meaning to mention, the parallelization would be good for more than just classifiers. We are likely to want to run cross-matching modules in parallel as well, and  classifiers may want to use results from more than one cross match.

troyraen:
This has sparked a good dialogue that has continued outside GitHub and is ongoing.

Issue #206, opened 2023-10-12
VM startup script exits due to bug, also unable to set VM metadata (main/elasticc branch)


hernandezc1:
At the beginning of `vm_startup.sh`, line 7 (`rm -rf "${brokerdir}"`) deletes the specified directory and its contents in order to start fresh. This is causing the startup script to exit after executing line 8: `cd "${workingdir}" || exit` because the directory `/home/broker` does not exist. See screenshot below:

![bug](https://github.com/mwvgroup/Pitt-Google-Broker/assets/83978600/859a39d2-16f1-4439-ad4c-4cedc6e8e7a6)

The source of the issue is that before PR #205  was merged into `main/elasticc`, the pitt-reader.user.keytab (`/home/broker/pitt-reader.user.keytab`) was made immutable. As a result, after running line 7, the directory `/home/broker` was not deleted because `pitt-reader.user.keytab` could not be removed. Since we've now deleted the `pitt-reader.user.keytab` file as a result of PR #205, the command in line 7 now deletes the directory.

After manually going through the rest of `vm_startup.sh`, I also noticed an error message when trying to set the VM metadata (lines 53-54). See the screenshot below:

![Screenshot 2023-10-12 at 6 25 46 PM](https://github.com/mwvgroup/Pitt-Google-Broker/assets/83978600/249d0633-b28a-41fe-9a14-7bd8a0faf068)

The message points me to two websites: [gcloud topic flags-file](https://cloud.google.com/sdk/gcloud/reference/topic/flags-file) and [gcloud topic escaping](https://cloud.google.com/sdk/gcloud/reference/topic/escaping). I tried implementing the suggestions provided in the second link, but I unfortunately kept getting the same error message.

I will continue to work on this issue tomorrow and provide additional updates.

troyraen:
Looks like there's really 2 issues here:

1. The script exits early because `workingdir` doesn't exist.
2. The syntax for setting VM instance metadata needs to be different when a key/value pair includes commas.

My thoughts:

1. Script exits early:
    - Move line 7 (`rm -rf "${brokerdir}"`) down to line 42 so that it happens right before the fresh files get downloaded.
    - Let's also avoid the `cd` altogether:
        - Delete line 8 (`cd "${workingdir}" || exit`)
        - Rename the `workingdir` to `consumerdir` for clarity (find and replace all occurrences in the file). 
[...]

PR #207, opened 2023-10-13
Fix bugs in vm_startup.sh

hernandezc1:
This PR addresses the two issues outlined in Issue #206:
1. The script exits early because `workingdir` doesn't exist
2. The syntax for setting VM instance metadata needs to be different when a key/value pair includes commas

Actions:
- [x] Rename the `workingdir` to `consumerdir` for clarity
- [x] Update the syntax for setting VM instance metadata
- [x] Move `rm -rf "${brokerdir}")` down in the script so that it happens right before the fresh files get downloaded
- [x] Re-create the directory `broker` (otherwise the fresh config files from the bucket cannot be downloaded)
- [x] Specify the full path to `fconfig` (otherwise the script is unable to set the Kafka topic & connector's configs: project and topics)

The motivation behind the last two items is discussed in Issue #206

### DIFF
--- a/broker/consumer/vm_startup.sh
+++ b/broker/consumer/vm_startup.sh
@@ -54,7 +54,7 @@ KAFKA_TOPIC="${KAFKA_TOPIC_FORCE:-${KAFKA_TOPIC_DEFAULT}}"
 PS_TOPIC="${PS_TOPIC_FORCE:-${PS_TOPIC_DEFAULT}}"
 # set VM metadata, just for clarity and easy viewing
 gcloud compute instances add-metadata "${consumerVM}" --zone "${zone}" \
-    --metadata ^:^CURRENT_PS_TOPIC=${PS_TOPIC}:CURRENT_KAFKA_TOPIC=${KAFKA_TOPIC}
+    --metadata "^:^CURRENT_PS_TOPIC=${PS_TOPIC}:CURRENT_KAFKA_TOPIC=${KAFKA_TOPIC}"
 
 #--- Set the connector's configs (project and topics)
 fconfig="/home/broker/consumer/ps-connector.properties"
--- a/broker/consumer/vm_startup.sh
+++ b/broker/consumer/vm_startup.sh
@@ -40,7 +40,7 @@ fi
 rm -rf "${brokerdir}"
 
 # create broker and consumer directories
-mkdir -p "${brokerdir}"
+mkdir "${brokerdir}"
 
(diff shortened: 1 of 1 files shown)


broker/consumer/vm_startup.sh
```diff
@@ -51,16 +54,16 @@ KAFKA_TOPIC="${KAFKA_TOPIC_FORCE:-${KAFKA_TOPIC_DEFAULT}}"
 PS_TOPIC="${PS_TOPIC_FORCE:-${PS_TOPIC_DEFAULT}}"
 # set VM metadata, just for clarity and easy viewing
 gcloud compute instances add-metadata "${consumerVM}" --zone "${zone}" \
-    --metadata "CURRENT_PS_TOPIC=${PS_TOPIC},CURRENT_KAFKA_TOPIC=${KAFKA_TOPIC}"
+    --metadata "^:^CURRENT_PS_TOPIC=${PS_TOPIC}:CURRENT_KAFKA_TOPIC=${KAFKA_TOPIC}"
 
 #--- Set the connector's configs (project and topics)
-fconfig="ps-connector.properties"
+fconfig="/home/broker/consumer/ps-connector.properties"
```
troyraen:
```suggestion
fconfig="${consumerdir}/ps-connector.properties"
```

Use the variable that's already defined.

broker/consumer/vm_startup.sh
```diff
@@ -51,16 +54,16 @@ KAFKA_TOPIC="${KAFKA_TOPIC_FORCE:-${KAFKA_TOPIC_DEFAULT}}"
...
 sed -i "s/PROJECT_ID/${PROJECT_ID}/g" "${fconfig}"
 sed -i "s/PS_TOPIC/${PS_TOPIC}/g" "${fconfig}"
 sed -i "s/KAFKA_TOPIC/${KAFKA_TOPIC}/g" "${fconfig}"
 
 #--- Set the Kafka offset
-fconfig="psconnect-worker.properties"
+fconfig="/home/broker/consumer/psconnect-worker.properties"
```
troyraen:
```suggestion
fconfig="${consumerdir}/psconnect-worker.properties"
```

troyraen (approved):
Two suggestions below. Otherwise looks good. Please test the script end to end to catch and fix any other bugs before merging.

hernandezc1:
Everything looks good. Script was tested end-to-end.

Issue #212, opened 2023-10-20
Defining multiple KAFKA topics that our consumer VM will listen to


hernandezc1:
In order to ingest alerts, we (or a user) must specify the `KAFKA_TOPIC` a consumer VM in our pipeline must listen to. PR #205 notes that this may be defined as a comma-separated list of topics, which allows a single consumer VM to listen to multiple topics simultaneously.

Currently, this is relevant in the context of the ELAsTiCC2 Challenge. The size of the alerts has increased significantly since the original ELAsTiCC Challenge, and as a result, DESC has decided to create three independent alert streams. For the recent test streams, this includes:
- `elasticc2-stN-wfd` -- WFD objects, including up to 365 days of previous sources and forced sources
- `elasticc2-stN-ddf-full` -- DDF objects, including up to 365 days of previous sources and forced sources
- `elasticc2-stN-ddf-limited` -- DDF objects, including up to 30 days of previous sources and forced sources

(Replace `N` with the numbers 1, 2, 3, ..., etc. These numbers represent the `N` th test stream before the re-start of the ELAsTiCC2 Challenge).

This issue serves to document my experience and findings associated with having our consumer VM listen to multiple KAFKA topics.

So far, I would like to note:
- If we define a comma-separated list of topics, and one of the topics is not live, the Kafka -> Pub/Sub Connector will allow alerts to be ingested from the topics that _are_ live.
- I opened up #213  to address a bug in the startup script of our consumer VM

troyraen:
> Even if one of the (multiple) topics is not live, the consumer will ingest alerts from the topics that are live

1. I think perhaps you meant that the _Kafka -> Pub/Sub Connector_ will behave this way if given a list of topics? There are certainly some consumer startup scripts in this repo that will not behave this way (in fact, some will just crash if given a list of topics).
2. What happens if the consumer starts ingesting the one live topic and then some time later the other topic goes live? Will it automatically start ingesting the new one as well? This is probably not high on our priority list, but will be important to learn at some point.

hernandezc1:
> 1. I think perhaps you meant that the _Kafka -> Pub/Sub Connector_ will behave this way if given a list of topics?

Yes, that's correct. I'll update the wording of the main comment.

hernandezc1:
[...]

PR #213, opened 2023-10-23
Fixes bug in consumer VM startup script

hernandezc1:
Related to Issue #212. 

For the ELAsTiCC2 Challenge, our consumer VM will be listening to multiple KAFKA topics simultaneously. When defining multiple topics, the variable `KAFKA_TOPIC` is initialized as a comma-separated list of topics: `KAFKA_TOPIC="topic1,topic2,etc"`. These topics appear in the file: `list.topics` in the following format.

```
topic1
topic2
etc
```

However, this creates a bug when the following code block from `vm_startup.sh` is executed.
```
if grep -Fq "${KAFKA_TOPIC}" "${fout_topics}"
    then
        alerts_flowing="true"  # start consuming
    else
        sleep 60s  # sleep 1 min, then try again
fi
```

This is because the `grep -Fq "${KAFKA_TOPIC}" "${fout_topics}"` command looks for the exact string, `topic1,topic2,etc`, in `list.topics`.

This PR updates the requirements needed to start the Kafka -> Pub/Sub connector for our consumer VM. In order to accommodate for multiple topics, we now parse through the string containing the comma-separated Kafka topics and append each topic name to an array called `KAFKA_TOPIC_ARRAY`. Using a `for` loop, we search to see whether the elements of `KAFKA_TOPIC_ARRAY` are present in `list.topics`. If a topic is present, then the value `true` is appended to separate array: `kafka_topic_present`.

If all of the topics are present in `list.topics`, then `KAFKA_TOPIC_ARRAY` and `kafka_topic_present` will have the same length. This requirement must now be satisfied in order to start the Kafka -> Pub/Sub connector.

The changes to the startup script have been successfully tested.

### DIFF
--- a/broker/consumer/vm_startup.sh
+++ b/broker/consumer/vm_startup.sh
@@ -87,12 +87,23 @@ do
     # which kills the while loop. no working suggestions found.
     # passing the error with `|| true`
 
+    
+    IFS=', ' read -r -a KAFKA_TOPIC_ARRAY <<< "$KAFKA_TOPIC"
+    KAFKA_TOPIC_PRESENT=()
+
     # check if our topic is in the list
-    if grep -Fq "${KAFKA_TOPIC}" "${fout_topics}"
-    then
-        alerts_flowing="true"  # start consuming
+    for topic in "${KAFKA_TOPIC_ARRAY[@]}"; 
+    do
+        if grep -Fq "${topic}" "${fout_topics}"
+        then
+            KAFKA_TOPIC_PRESENT+=("True")
+        fi
+    done
+    
+    if [ ${#KAFKA_TOPIC_PRESENT[@]} -eq ${#KAFKA_TOPIC_ARRAY[@]} ]; then
+        alerts_flowing="true"
     else
-        sleep 60s  # sleep 1 min, then try again
+        sleep 60s
     fi
 done
 
--- a/broker/consumer/vm_startup.sh
+++ b/broker/consumer/vm_startup.sh
@@ -46,12 +46,12 @@ mkdir "${brokerdir}"
 gsutil -m cp -r "gs://${broker_bucket}/consumer" "${brokerdir}"
 gsutil -m cp -r "gs://${broker_bucket}/schema_maps" "${brokerdir}"
 
-#--- Set default Kafka topic
-KAFKA_TOPIC_DEFAULT="elasticc-2022fall"
-
 #--- Set the topic names to the "FORCE" metadata attributes if exist, else defaults
+KAFKA_TOPIC_DEFAULT="elasticc-2022fall" # set default Kafka topic
 KAFKA_TOPIC="${KAFKA_TOPIC_FORCE:-${KAFKA_TOPIC_DEFAULT}}"
 PS_TOPIC="${PS_TOPIC_FORCE:-${PS_TOPIC_DEFAULT}}"
+# append the Kafka topic name(s) into an array by using a comma as the delimiter
+IFS=',' read -r -a KAFKA_TOPIC_ARRAY <<< "$KAFKA_TOPIC"
 # set VM metadata, just for clarity and easy viewing
 gcloud compute instances add-metadata "${consumerVM}" --zone "${zone}" \
     --metadata "^:^CURRENT_PS_TOPIC=${PS_TOPIC}:CURRENT_KAFKA_TOPIC=${KAFKA_TOPIC}"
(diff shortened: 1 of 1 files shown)


broker/consumer/vm_startup.sh
```diff
@@ -87,17 +87,28 @@ do
     # which kills the while loop. no working suggestions found.
     # passing the error with `|| true`
 
+    
+    IFS=', ' read -r -a KAFKA_TOPIC_ARRAY <<< "$KAFKA_TOPIC"
```
troyraen: This should go in the section at the top of the file that defines topic-related variables. Actually there's currently two sections, starting on lines 49 and 52, I would combine them into one and add this there.

Add a comment describing what this line does. Shell is not as intuitive to read as python, so more comments are helpful.

broker/consumer/vm_startup.sh
```diff
@@ -87,17 +87,28 @@ do
...
+    KAFKA_TOPIC_PRESENT=()
```
troyraen: It is important that this variable  be defined right before the next for `loop` starts. Make this more clear to future readers by moving this down between the current lines 94 and 95 so that it's actually part of the same block of code. 

Additionally, using all caps to name a variable is typically only done for environment variables and module level variables. This is neither, so use snake case.

broker/consumer/vm_startup.sh
```diff
@@ -87,17 +87,28 @@ do
...
+
     # check if our topic is in the list
```
troyraen: Update this comment to better describe what will happen in this code block. Include info about different cases, e.g., multiple topics but one goes live before the other. You may want to point at issue #212 where the  ongoing discussion will occur and options & unknowns can be described in more detail.

broker/consumer/vm_startup.sh
```diff
@@ -87,17 +87,28 @@ do
...
+    for topic in "${KAFKA_TOPIC_ARRAY[@]}"; 
+    do
+        if grep -Fq "${topic}" "${fout_topics}"
+        then
+            KAFKA_TOPIC_PRESENT+=("True")
+        fi
+    done
+    
+    if [ ${#KAFKA_TOPIC_PRESENT[@]} -eq ${#KAFKA_TOPIC_ARRAY[@]} ]; then
+        alerts_flowing="true"
```
troyraen: Add a comment to the first line explaining what it does. I imagine it's something like, "check whether the two arrays are the same length".

If I understand correctly, this will cause the Connector to start up only if all topics are present. If so:
1. This should be explicitly stated in the comment above the code block (line 94).
2. In #212 you said:

    > Even if one of the (multiple) topics is not live, the consumer will ingest alerts from the topics that are live.

    I'm guessing that what you actually tested was the behavior of the _Kafka -> Pub/Sub Connector_ (launched by the last command in the script `/bin/connect-standalone ...`). I recommend updating the languaging in the issue to reflect that.

broker/consumer/vm_startup.sh
```diff
@@ -87,17 +87,28 @@ do
...
     else
-        sleep 60s  # sleep 1 min, then try again
+        sleep 60s
```
troyraen: Why delete the comment?

broker/consumer/vm_startup.sh
```diff
@@ -87,17 +87,28 @@ do
...
     fi
 done
 
 #--- Start the Kafka -> Pub/Sub connector, save stdout and stderr to file
 /bin/connect-standalone \
     "${consumerdir}/psconnect-worker.properties" \
     "${consumerdir}/ps-connector.properties" \
-    &>> "${fout_run}"
+    &>> "${fout_run}"
```
troyraen, line 117: This is not a real change.

troyraen (changes requested):
Some changes requested below, mostly related to code organization and comments.

One overall question: Does this still work if `KAFKA_TOPIC` is a single topic? We don't want to lose the ability to do that.

BTW, I didn't actually test this and I don't plan to. If I haven't said it explicitly before, I am no longer testing all of your PRs, especially the ones that affect code that only you are using right now. I'm reviewing it as best I can from reading it. It's up to you to try to think through all the possibilities when testing make sure your stuff actually works.

broker/consumer/vm_startup.sh
hernandezc1: my mistake, this shouldn't have been deleted

broker/consumer/vm_startup.sh
hernandezc1: I've noticed that this has been happening more frequently in a few different PRs. I'll make sure to review my changes before committing.

hernandezc1: Do you have any suggestions on how to fix this?

hernandezc1:
> One overall question: Does this still work if `KAFKA_TOPIC` is a single topic? We don't want to lose the ability to do that.

Yes, it still works.
 
> BTW, I didn't actually test this and I don't plan to. If I haven't said it explicitly before, I am no longer testing all of your PRs, especially the ones that affect code that only you are using right now. I'm reviewing it as best I can from reading it. It's up to you to try to think through all the possibilities when testing make sure your stuff actually works.

I'll request your review again momentarily. I have been testing all of my changes thoroughly. I plan on running additional tests for this PR in the next few days, just for my own sanity check.

broker/consumer/vm_startup.sh
hernandezc1: I added a more descriptive comment, please let me know what you think. I felt it might be more appropriate to keep the discussion about the different cases in Issue #212, but if you disagree, let me know and I will update the comment again

broker/consumer/vm_startup.sh
hernandezc1: >    I'm guessing that what you actually tested was the behavior of the _Kafka -> Pub/Sub Connector_ (launched by the last command in the script `/bin/connect-standalone ...`). I recommend updating the languaging in the issue to reflect that.

Yes, this is correct. I will update the main comment shortly

troyraen (approved):
LGTM

Issue #121, opened 2021-12-15
Connect to LIGO/Virgo stream


troyraen:
Related to #126 

LIGO/Virgo alerts are distributed through [GCN](https://gcn.gsfc.nasa.gov/) in VOEvent format, XML documents.

- [LIGO/Virgo Public Alerts User Guide](https://emfollow.docs.ligo.org/userguide/index.html)
- [Instructions for a Python listener](https://emfollow.docs.ligo.org/userguide/tutorial/index.html) (`PyGCN`)
- C listener [`voevent_client_demo.c`](https://gcn.gsfc.nasa.gov/gcn/voevent_client_demo.c) --- demonstrates a VOEvent socket connection to GCN/TAN (we probably want the Python listener instead, but leaving this link for reference)

wmwv:
GCN is runs a Kafka server and supply alerts for LIGO, IceCube, Swift, Fermi, ...

Data rate is ~10 / day.  (astrophysical events; ignore Swift slews)

What's the next step to get things set up?

troyraen:
Next steps would be to connect our consumer to their Kafka server, then update our pipeline  to handle those alerts. It would be helpful if I merged #123 and #194. I will move them up on my priority list.

troyraen:
#123 is now merged and #194 is ready for review.

hernandezc1:
There are a few prerequisites that need to be satisfied in order to receive LIGO/Virgo/KAGRA alerts from [GCN Notices](https://gcn.nasa.gov/docs/notices) over Kafka. A user must first be authenticated, and this can be done by [signing up](https://gcn.nasa.gov/quickstart) for an account using the link provided. An authenticated user is then given a `client ID` and `client_secret` value, as well as a sample Python script that is used to receive and parse through notices. 

Users can select to receive **either** JSON-serialized or Avro-serialized notices over Kafka (these two options are recommended over receiving [VOEvent Notices via GCN Classic](https://emfollow.docs.ligo.org/userguide/tutorial/receiving/classic.html)). In order to consume notices over Kafka, a user will need to select one of the two options:
1. [gcn-kafka](https://pypi.org/project/gcn-kafka/) for connecting to [GCN Kafka](https://gcn.nasa.gov/) to receive JSON-serialized notices over Kafka
[...]

PR #215, opened 2023-11-27
Set up configuration files to connect to LIGO/Virgo/KAGRA alert stream

hernandezc1:
Related to #121 

This PR creates new directories within `broker/consumer/` and `broker/setup_broker/` specific to LIGO/Virgo/KAGRA (LVK). Within these directories are scripts and/or configuration files that are used to deploy and [authenticate a consumer VM](https://docs.confluent.io/platform/current/kafka/authentication_sasl/authentication_sasl_oauth.html#configuring-ak-clients) to connect to the LVK gravitational wave alert stream. The authentication process uses the [OAuth 2 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749).

These changes have been successfully tested.


hernandezc1:
Here are a list of websites that I've used to learn more about this process.

1. [Configuring OAuth2 authentication for Kafka clients](https://docs.cloudera.com/runtime/7.2.15/kafka-securing/topics/kafka-securing-oauth-clients.html)
2. [Configuring Kafka Clients](https://docs.confluent.io/platform/current/kafka/authentication_sasl/authentication_sasl_oauth.html#configuring-ak-clients)
3. [Class OAuthBearerLoginModule](https://kafka.apache.org/20/javadoc/org/apache/kafka/common/security/oauthbearer/OAuthBearerLoginModule.html)

The last link discusses the need to implement an interface [AuthenticateCallbackHandler](https://kafka.apache.org/20/javadoc/org/apache/kafka/common/security/auth/AuthenticateCallbackHandler.html) class. The `sasl.login.callback.handler.class` property allows you to specify the Java class that will manage these SASL login callbacks. This class is responsible for handling and responding to the challenges presented during the authentication process.

This is what I was referring to yesterday (@troyraen) when I mentioned that I thought I'd have to write my own code.

broker/consumer/ligo/consumer.properties
troyraen, line 1: I'm not sure how you're calling these files, since you haven't updated the consumer startup script. But I think this file is misnamed and should be `admin.properties` like for ZTF and Rubin.

broker/consumer/ligo/consumer.properties
```diff
@@ -0,0 +1,11 @@
+# see https://docs.confluent.io/platform/current/kafka/authentication_sasl/authentication_sasl_oauth.html#configuring-ak-clients
```
troyraen: Re: this and the links you posted in your comment above... Be careful about which docs you're looking at. For example, this link is not wrong w.r.t. the OAuth settings needed here, but it's also not the actual specification docs for this file. Look at the link in our other admin.properties files to understand what configs are allowed here. Also, if you haven't yet:

- Learn the difference between Confluent Kafka, Kafka Connect, Kafka Brokers and Consumers, etc. They are interdependent, and if you don't understand the relationships it's easy to get confused when troubleshooting and reading docs.
- Read the [Kafka primer](https://pitt-broker.readthedocs.io/en/latest/broker/primers-for-developers/kafka-console-connect.html) in our docs, and especially the section about Kafka Connectors.

broker/consumer/ligo/consumer.properties
```diff
@@ -0,0 +1,11 @@
...
+
+bootstrap.servers=kafka.gcn.nasa.gov:9092
+security.protocol=SASL_SSL
+sasl.mechanism=OAUTHBEARER
+sasl.oauthbearer.token.endpoint.url=https://auth.gcn.nasa.gov/oauth2/token
+sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
+     unsecuredLoginStringClaim_sub="pitt-google-broker-test" \
+     oauth.client.id="CLIENT_ID" \
+     oauth.client.secret="CLIENT_SECRET" \
+     scope="gcn.nasa.gov/kafka-public-consumer";
```
troyraen: This `OAuthBearerLoginModule` creates unsecured tokens -- it does not accept the client ID, secret, and scope options. (See your link above for the list of options that it does take.)

Once you get to the point where you *can* plug in those options, it will be important to see the code you're using to substitute values for the `CLIENT_ID` and `CLIENT_SECRET`. That is an operation where things can easily go wrong.

broker/consumer/ligo/psconnect-worker.properties
```diff
@@ -0,0 +1,46 @@
...
+value.converter=org.apache.kafka.connect.converters.ByteArrayConverter
+
+# workers need to use SASL
+sasl.mechanism=OAUTHBEARER
+security.protocol=SASL_SSL
+sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
+     unsecuredLoginStringClaim_sub="pittgoogle-broker"\
+     oauth.client.id="CLIENT_ID"\
+     oauth.client.secret="CLIENT_SECRET"\ 
+     oauth.token.endpoint.url="https://auth.gcn.nasa.gov/oauth2/token";
```
troyraen: I recall having trouble getting these config files to work because of formatting and whitespace issues.

- I would put a space before the trailing `\` here and below (though I see one instance of this in the Rubin files, so either that's wrong or this is fine).
- In particular, the coloring of `oauth.token.endpoint.url` vs the other keys makes me think there is a formatting problem here.
- Also, I would check to make sure there is no trailing whitespace at the end of a line in any of these files.

broker/consumer/ligo/psconnect-worker.properties
```diff
@@ -0,0 +1,46 @@
...
+
+# settings with `consumer` prefixes are passed through to the Kafka consumer
+# consumer.group.id must begin with our username: pittgoogle-idfint
+consumer.group.id=pittgoogle-idfint-kafka-pubsub-connector
```
troyraen: The username referenced here is our Rubin username. Sometimes the client can set whatever group ID it wants (and in that case this value would still work for GCN), but sometimes the broker puts restrictions on it. So if you get to the point where the auth works but you still can't get fully connected, check with GCN about this.

troyraen (commented):
Looks like you’re right — there’s no built-in way to handle the OAuth with a client ID and secret. Here are some things you can try:

- Reach out to whoever manages the GCN Kafka stream and ask if there is any other way that we can authenticate besides OAuth. They may be able to set us up with a keytab (like ZTF) or something else. IMO, this is by far the most preferable path forward.
- This [Medium article](https://3bit-techs.medium.com/confluent-kafka-security-with-oauthbearer-9b8cf3c7c784) appears to contain the Java code we'd need for custom OAuth handlers. Maybe we'll get lucky and can simply copy/paste it in here and use it, but I doubt it. IMO, this is *not* the direction to go in. But you can try it if you want. (Perhaps Ross can help, especially with the Java.)
- You could try setting up an entirely new consumer module using the [GCN python client](https://gcn.nasa.gov/docs/client#python). This may be a quicker way to get connected and start playing with GW alerts. Longer term, I would not recommend a python client like this for the larger ZTF and LSST streams, but this might actually be a fine way for us to connect to GCN given the very small size of the stream.

I'm pretty sure that the lack of a secured OAuth handler class in Kafka Connect is the fatal problem right now, though I do see a few other things in this PR that may cause problems. I have commented on them below so that you are aware of them for the future.

hernandezc1:
After reaching out to GCN support, their documentation on the Kafka Client Setup was updated to include instructions for the official Kafka command line tools which use Java and come with either [Apache Kafka](https://kafka.apache.org/documentation/#quickstart) or [Confluent](https://docs.confluent.io/kafka/operations-tools/kafka-tools.html). The documentation can be accessed by clicking this [link](https://gcn.nasa.gov/docs/client#java).

After updating the `consumer.properties` file to reflect what's stated in the instructions, I ssh'd into our consumer's VM and ran the following command:

```
/bin/kafka-topics \
--bootstrap-server kafka.gcn.nasa.gov:9092 \
--command-config /home/broker/consumer/ligo/consumer.properties \
--list
```

This gave me the error:
`Exception in thread "main" org.apache.kafka.common.config.ConfigException: Invalid value org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginCallbackHandler for configuration sasl.login.callback.handler.class: Class org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginCallbackHandler could not be found.
        at org.apache.kafka.common.config.ConfigDef.parseType(ConfigDef.java:729)
        at org.apache.kafka.common.config.ConfigDef.parseValue(ConfigDef.java:475)
        at org.apache.kafka.common.config.ConfigDef.parse(ConfigDef.java:468)
        at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:108)
        at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:142)
        at org.apache.kafka.clients.admin.AdminClientConfig.<init>(AdminClientConfig.java:217)
        at org.apache.kafka.clients.admin.Admin.create(Admin.java:63)
        at kafka.admin.TopicCommand$AdminClientTopicService$.createAdminClient(TopicCommand.scala:216)
        at kafka.admin.TopicCommand$AdminClientTopicService$.apply(TopicCommand.scala:220)
        at kafka.admin.TopicCommand$.main(TopicCommand.scala:57)
        at kafka.admin.TopicCommand.main(TopicCommand.scala)`

@troyraen is this an error message you've seen before?

hernandezc1:
I noticed that there were slight discrepancies between the sample code provided in the GCN documentation and the instructions provided on the Confluent Documentation (see [Configure OAuth Authentication for Kafka Clients](https://docs.confluent.io/platform/current/kafka/authentication_sasl/authentication_sasl_oauth.html#configure-ak-clients)). After updating consumer.properties once again, and entering the same command, I get the same error message.

It seems to me that either the class is missing from the classpath or it's not available in the Kafka libraries we're using.

broker/consumer/ligo/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>io.pitt-broker</groupId>
+    <artifactId>lvk-stream</artifactId>
+    <version>0.1.0</version>
+    
+    <repositories>
+
+      <repository>
```
jrossthomson: Try adding the repository indicate here: 

https://github.com/jairsjunior/kafka-oauth?tab=readme-ov-file#kafka-client-configuration-producerconsumer

broker/consumer/vm_install.sh
```diff
@@ -74,4 +74,4 @@ startupscript="gs://${broker_bucket}/consumer/vm_startup.sh"
 gcloud compute instances add-metadata "$consumerVM" --zone "$zone" \
     --metadata startup-script-url="$startupscript"
 echo "vm_install.sh is complete. Shutting down."
-shutdown -h now
+shutdown -h now
```
hernandezc1, line 77: I previously edited this script and pushed the changes to this PR. Eventually, I decided to discard those changes by replacing the changed code with the original code. Because of this, Github is marking the last line of this file and `broker/consumer/vm_startup.sh` as being changed (I did the same thing with `broker/consumer/vm_startup.sh`), despite the code being identical. I haven't figured out a way to resolve this

wmwv: It's because the `master` has a line feed character at the end, and the changed version doesn't.  The correct answer is that it should have a line feed character.

wmwv: The red circle with the line through it is GitHubs way of telling you this.

wmwv: You will want to configure your default editor to make sure there is always a line feed character at the end of your files.

wmwv (approved):
I've read it but not run it myself.  Go ahead and merge and deploy.  Keep an eye out to see if there are any quick fixes needed once we have this running for a bit.

broker/consumer/lvk/ps-connector.properties
```diff
@@ -0,0 +1,32 @@
...
+# Unique name for the Pub/Sub sink connector.
+name=CPSSinkConnector
+# Tha Java class for the Pub/Sub sink connector.
+connector.class=com.google.pubsub.kafka.sink.CloudPubSubSinkConnector
+# The maximum number of tasks that should be created for this connector.
+tasks.max=10
+# Set the key converter for the Pub/Sub sink connector.
+key.converter=org.apache.kafka.connect.storage.StringConverter
+# Set the value converter for the Pub/Sub sink connector.
+value.converter=org.apache.kafka.connect.converters.ByteArrayConverter
```
troyraen: I think these two entries don't belong in this file. If they do need to be here for some reason, the values should match those in psconnect-worker-authenticated.properties. In particular, I would always choose `ByteArrayConverter` when possible so that we do not alter the data when ingesting.

broker/consumer/lvk/vm_startup.sh
```diff
@@ -0,0 +1,104 @@
...
+gcloud compute instances add-metadata "$consumerVM" --zone "$zone" \
+    --metadata="PS_TOPIC=${PS_TOPIC},KAFKA_TOPIC=${KAFKA_TOPIC}"
+
+#--- Files this script will write
+fout_run="${workingdir}/run-connector.out"
+fout_topics="${workingdir}/list.topics"
+
+#--- Set the connector's configs (client ID, client secret, project, and topics)
+# define LVK-related parameters
+surveydir="${workingdir}/lvk"
```
troyraen: Two recommendations here:
1. Use the `survey` variable rather than hard-coding the value here.
2. Just make this the `workingdir` so there's no need for a separate `surveydir`. In particular, it would be better for the output files to be together with the other files for LVK instead of up one level with the ZTF files.

broker/setup_broker/lvk/create_vms.sh
```diff
@@ -0,0 +1,51 @@
...
+    googlelogging="google-logging-enabled=true"
+    startupscript="startup-script-url=gs://${broker_bucket}/consumer/${survey}/vm_install.sh"
+    shutdownscript="shutdown-script-url=gs://${broker_bucket}/consumer/${survey}/vm_shutdown.sh"
+    gcloud compute instances create "$consumerVM" \
+        --zone="$zone" \
+        --address="$consumerIP" \
+        --machine-type="$machinetype" \
+        --scopes=cloud-platform \
+        --metadata="${googlelogging},${startupscript},${shutdownscript}" \
+        --tags=ztfport # for the firewall rule to open the port
```
troyraen: If this is actually needed, use the `lvkport` that you're creating in setup_broker.sh.

broker/setup_broker/lvk/create_vms.sh
```diff
@@ -0,0 +1,51 @@
...
+    if [ "$testid" != "False" ]; then
+        gcloud compute instances delete "$consumerVM" --zone="$zone"
+        gcloud compute addresses delete "${consumerIP}" --region="${region}"
+    fi
+
+#--- Create resources
+else
+#--- Consumer VM
+    # create a static ip address so that it can be whitelisted by the survey
+    gcloud compute addresses create "${consumerIP}" --region="${region}"
```
troyraen: Is a static IP really needed for LVK? If not, then might as well make the script simpler by getting rid of it. I don't think it's doing any harm, so ok to leave as-is if you want, but then at least update the comment so that it doesn't seem like the IP needs to be whitelisted for this to work.

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -0,0 +1,92 @@
...
+
+#--- Create a firewall rule to open the port used by Kafka/ZTF
+# on any instance with the flag --tags=ztfport
+    echo
+    echo "Configuring LVK/Kafka firewall rule..."
+    gcloud compute firewall-rules create 'lvkport' \
+        --allow=tcp:9094 \
+        --description="Allow incoming traffic on TCP port 9094" \
+        --direction=INGRESS \
+        --enable-logging
```
troyraen: I don't think you're using port 9094 at all, so I don't think you need this.

broker/setup_broker/lvk/create_vms.sh
```diff
@@ -0,0 +1,51 @@
...
+    # create VM
+    machinetype=e2-standard-2
```
troyraen, line 33: The install probably requires a machine this big, but after that it should be able to run on a much smaller (and cheaper) machine. For now, I recommend changing the machine size manually after the install. As I recall, I haven't figure out a way to put that into these scripts since the machine has to be shutdown before changing the size.

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -0,0 +1,92 @@
...
+    gcloud pubsub topics create "${topic_alerts}"
+
+    # Set IAM policies on resources
+    user="allUsers"
+    roleid="projects/${GOOGLE_CLOUD_PROJECT}/roles/userPublic"
+    gcloud pubsub topics add-iam-policy-binding "${topic_alerts}" --member="${user}" --role="${roleid}"
+    user="542312691208-compute@developer.gserviceaccount.com"
+    roleid="roles/secretmanager.secretAccessor"
+    gcloud secrets add-iam-policy-binding "${client_id}" --member="serviceAccount:${user}" --role="${roleid}"
+    gcloud secrets add-iam-policy-binding "${client_secret}" --member="serviceAccount:${user}" --role="${roleid}"
```
troyraen: I think it's probably better to put this in the same place as where the secrets are actually created, which is not in this script.

troyraen (approved):
A few recommendations below. Congrats on getting this working!

PR #216, opened 2024-01-31
Update to Python 312 in cloud function deploy

wmwv:
Update the `deploy.sh` scripts to Python 3.12 `--runtime python312`

Draft PR.  Need to test.


troyraen:
If you haven't tested this locally yet, we should. The modules install `pgb-broker-utils` from PyPI (even though the source code is in broker/broker_utils... it will be removed from this repo once we switch to `pittgoogle-client`). If the PyPI version is not compatible with python==3.12, the modules will fail. This may be a good job for @hernandezc1 .

- create or activate a python 3.12 conda environment
- pip-install `pgb-broker-utils`
- pull an alert from the `ztf-loop` Pub/Sub stream
- test the modules

troyraen:
Once we get to the stage of testing in the cloud, here are the basic instructions:

- Follow [Setup the Broker](https://pitt-broker.readthedocs.io/en/latest/broker/run-a-broker-instance/setup-broker.html) (ignore "Upload Kafka Authentication Files" -- we don't need the consumer for this). 
    - Choose and use a `testid`.
    - Use `version="4.02"` (ZTF schema version).
- The `ps_to_gcs/` cloud function (called "upload_bytes_to_bucket" in the GCP console) sits between the consumer and the rest of the pipeline. So publish alerts to its trigger topic and they will (should) make their way through the whole pipeline. Do this by either: (1) manually pulling messages from the `ztf-loop` pubsub stream and publishing them to the trigger topic; or (2) by switching the trigger topic to point at the `ztf-loop` stream (either in code before deploying, or I think this is possible from the GCP console after deploying). Option (2) requires that the test deployment is in the same GCP project as the ZTF broker because GCP doesn't allow you to connect the kind of Cloud Functions we use to external projects.
- Check whether the expected data is in the output Pub/Sub topics, BigQuery tables, and GCS buckets. Also check the Cloud Functions metrics and logs. See [View and Access Resources](https://pitt-broker.readthedocs.io/en/latest/broker/run-a-broker-instance/view-resources.html) for instructions on how to find those things.

troyraen:
I did some local testing and I think the additional changes we need will be minimal. I added them in #219. If we can merge those in here, then I think this is ready for cloud testing.

broker/cloud_functions/check_cue_response/deploy.sh
troyraen, line 1: I think we can ignore this cloud function here. I'd like to just delete it. It used to do several checks on whether things were running properly and generate a notification/alarm if they weren't. But most (I think all) have become unnecessary over time. Some kind of check/alarm system might be useful, but I don't think there's anything about this one that we should salvage.

troyraen:
@wmwv will you also delete the file `broker/broker_utils/requirements_pinned.txt` in this PR? It is causing a bunch of dependabot security warnings for this repo. I don't think we need to make changes to broker_utils for python 3.12, so no good reason to keep this file around anymore.

hernandezc1 (approved):

PR #231, opened 2024-06-07
Update Rubin ps-connector.properties file

hernandezc1:
`sed -i "s/PS_TOPIC/${PS_TOPIC}/g" ${fconfig}`

Line 71 in the VM's startup script, displayed above, updates the value `PS_TOPIC` in the ps-connector.properties file with the appropriate PubSub topic name. Due to a typo in the ps-connector.properties file, the startup script was unable to update the `PS_TOPIC` value. This PR fixes the typo.

Changes have been tested. No additional bugs detected

### DIFF
--- a/broker/consumer/rubin/ps-connector.properties
+++ b/broker/consumer/rubin/ps-connector.properties
@@ -12,7 +12,7 @@ tasks.max=1
 # set kafka the topic
 topics=KAFKA_TOPIC
 # set the Pub/Sub configs
-cps.topic=PUBSUB_TOPIC
+cps.topic=PS_TOPIC
 cps.project=PROJECT_ID
 # include Kafka topic, partition, offset, timestamp as msg attributes
 metadata.publish=true


wmwv (approved):
This change is obviously a good fix.

wmwv:
But more broadly, this sed command implies that PS_TOPIC can't have special characters (~, &, \1) that would might something in a regex or be interpreted by the sed command (/).  Is that already a requirement by GCP, and is it already documented?

hernandezc1:
> But more broadly, this sed command implies that PS_TOPIC can't have special characters (~, &, \1) that would might something in a regex or be interpreted by the sed command (/). Is that already a requirement by GCP, and is it already documented?

@wmwv GCP does have a [guideline for naming a topic, subscription, schema, or snapshot](https://cloud.google.com/pubsub/docs/pubsub-basics#resource_names). With the way Pub/Sub resources are currently created, I can see this becoming an issue if the `testid` contains a special character. However, [our docs](https://pitt-broker.readthedocs.io/en/u-tjr-update_broker_utils/broker/broker-instance-keywords.html#keywords) explicitly state that the `testid` is a string of lowercase letters and numbers, beginning with a letter.

PR #232, opened 2024-06-12
Store LVK alerts in BigQuery

hernandezc1:
This PR creates the necessary files and subdirectories needed to store LIGO/Virgo/KAGRA (LVK) alert information in BigQuery.

Here is a more in-depth explanation of how things work.

**setup_broker dir:**
The `setup_broker/lvk` directory contains several bash scripts that are used to deploy an instance of our broker that is specific to LVK. In this PR, I've updated the "main" deployment script (`setup_broker.sh`) so that it defines and creates (or deletes) the necessary BigQuery and Pub/Sub resources to store LVK alert information in BigQuery. 

The idea here is that LVK alerts that our consumer publishes to a Pub/Sub topic (`lvk-alerts`) will trigger our Cloud Function (`lvk-storeBigQuery`), and the Cloud Function will insert the alert data into a BigQuery table and publish a message to a new Pub/Sub topic (`lvk-BigQuery`). The message published to `lvk-BigQuery` contains the original alert information as well as the alert type and the name of the table to which the alert information was stored to as message attributes.

In `setup_broker/lvk` you'll also see that I've added a `template` directory, which contains the schema of the BigQuery table.

**cloud_functions dir**
The `cloud_functions/lvk` directory contains four new files: a `README`, a `requirements.txt` file, the cloud function's deployment script (`deploy.sh`) and a Python script (`main.py`) used to store the alert information in BigQuery and publish a message to the topic `lvk-BigQuery`. `main.py` now uses the [`pittgoogle-client`](https://mwvgroup.github.io/pittgoogle-client/).


broker/setup_broker/templates/bq_lvk_alerts_schema.json
troyraen, line 1: My vote is to put this (and the other LVK templates) down a level in the LVK subdirectory.

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+logger = logging_client.logger(log_name)
+
+# GCP resources used in this module
+pittgoogle_project = pittgoogle.ProjectIds().pittgoogle
+MODULE_NAME = f"alerts_{VERSIONTAG}"
+TABLE = pittgoogle.Table.from_cloud(MODULE_NAME, survey=SURVEY, testid=TESTID)
+TOPIC = pittgoogle.Topic.from_cloud("BigQuery", survey=SURVEY, testid=TESTID, projectid=PROJECT_ID)
+
+# 'context' is an unused argument in the function below that is required
+# see https://cloud.google.com/functions/1stgendocs/writing/write-event-driven-functions#background-functions
```
troyraen: Move this inside the docstring

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+def run(event: dict, _context: functions_v1.context.Context) -> None:
```
troyraen, line 26: No need for the underscore. Let's leave these named as they are in Google's docs.

```suggestion
def run(event: dict, context: functions_v1.context.Context) -> None:
```

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+        table_dict = {"alerts_table": None}
+
+    return table_dict
+
+def _create_outgoing_alert(alert: pittgoogle.alert.Alert, table_dict: dict) -> pittgoogle.alert.Alert:
+    """Create an announcement of the table storage operation to Pub/Sub."""
+    # collect attributes
+    attrs = {
+        "alert_table": table_dict['alerts_table'],
+        "type": alert.dict['alert_type']
```
troyraen: This field comes directly from the alert. Don't change the name.

```suggestion
        "alert_type": alert.dict['alert_type'],
```

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+        msg = f"Error inserting to alerts table: {errors}"
+        logger.log_text(msg, severity="DEBUG")
+        table_dict = {"alerts_table": None}
+
+    return table_dict
+
+def _create_outgoing_alert(alert: pittgoogle.alert.Alert, table_dict: dict) -> pittgoogle.alert.Alert:
+    """Create an announcement of the table storage operation to Pub/Sub."""
+    # collect attributes
+    attrs = {
```
troyraen, line 79: Pass on the incoming alert attributes.

```suggestion
    attrs = {
        **alert.attributes,
```

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+
+# connect to the cloud logger
+log_name = "store-bigquery-cloudfnc"  # same log for all broker instances
+logging_client = logging.Client()
+logger = logging_client.logger(log_name)
+
+# GCP resources used in this module
+pittgoogle_project = pittgoogle.ProjectIds().pittgoogle
+MODULE_NAME = f"alerts_{VERSIONTAG}"
+TABLE = pittgoogle.Table.from_cloud(MODULE_NAME, survey=SURVEY, testid=TESTID)
```
troyraen: This module is not called "alerts". There's no need for a `MODULE_NAME` variable here so just skip it.

```suggestion
TABLE = pittgoogle.Table.from_cloud(f"alerts_{VERSIONTAG}", survey=SURVEY, testid=TESTID)
```

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+TESTID = os.getenv("TESTID")
+VERSIONTAG = os.getenv("VERSIONTAG")
+
+# connect to the cloud logger
+log_name = "store-bigquery-cloudfnc"  # same log for all broker instances
+logging_client = logging.Client()
+logger = logging_client.logger(log_name)
+
+# GCP resources used in this module
+pittgoogle_project = pittgoogle.ProjectIds().pittgoogle
```
troyraen: This is not being used anywhere. 

```suggestion
```

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -51,16 +69,19 @@ if [ "${teardown}" != "True" ]; then
...
     roleid="projects/${GOOGLE_CLOUD_PROJECT}/roles/userPublic"
     gcloud pubsub topics add-iam-policy-binding "${topic_alerts}" --member="${user}" --role="${roleid}"
+    gcloud pubsub topics add-iam-policy-binding "${topic_storebigquery}" --member="${user}" --role="${roleid}"
 
 else
     # ensure that we do not teardown production resources
     if [ "${testid}" != "False" ]; then
         o="GSUtil:parallel_process_count=1" # disable multiprocessing for Macs
+        bq rm -r -f "${PROJECT_ID}:${bq_dataset}"
         gsutil -m -o "${o}" rm -r "gs://${broker_bucket}"
```
troyraen: The `o` variable goes with the `gsutil` command, so don't separate them.

```suggestion
        o="GSUtil:parallel_process_count=1" # disable multiprocessing for Macs
        gsutil -m -o "${o}" rm -r "gs://${broker_bucket}"
        bq rm -r -f "${PROJECT_ID}:${bq_dataset}"
```

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -51,16 +69,19 @@ if [ "${teardown}" != "True" ]; then
...
         gcloud pubsub topics delete "${topic_alerts}"
```
troyraen, line 88: Please also delete the other Pub/Sub resource.

```suggestion
        gcloud pubsub topics delete "${topic_alerts}"
        gcloud pubsub topics delete "${topic_storebigquery}"
```

broker/cloud_functions/lvk/store_BigQuery/deploy.sh
```diff
@@ -0,0 +1,42 @@
...
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
+survey="${3:-lvk}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-O4}"
+
+#--- GCP resources used in this script
+store_bq_trigger_topic="${survey}-alerts"
+store_bq_CF_name="${survey}-store_in_BigQuery"
```
troyraen: Let's use lower case for all resource names.

```suggestion
store_bq_CF_name="${survey}-store_in_bigquery"
```

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -8,7 +8,9 @@ teardown="${2:-False}"
 # "True" tearsdown/deletes resources, else setup
 survey="${3:-lvk}"
 # name of the survey this broker instance will ingest
-region="${4:-us-central1}"
+observation_run="${4:-4}"
```
troyraen: Leave this variable named `schema_version` like we do for other surveys. That's what this variable really is and how it's used in the code. This will need to be changed in more places as well.

```suggestion
schema_version="${4:-4}"
```

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -8,7 +8,9 @@ teardown="${2:-False}"
...
+versiontag="O${observation_run}" # 4 -> O4
```
troyraen: This line should prepend a "v" for "version". Does LVK call this schema version "O4"? If so, that is the value that should be passed in as `schema_version` by the user. This line of code should convert that to "vO4" so that we end up with a table named with the syntax `<survey>.alerts_v<schema_version>`. This probably needs to be changed in other places as well.

```suggestion
versiontag="v${observation_run}" # 4 -> v4
```

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -34,15 +36,31 @@ fi
 
 #--- GCP resources used directly in this script
 broker_bucket="${PROJECT_ID}-${survey}-broker_files"
+bq_dataset="${survey}"
 topic_alerts="${survey}-alerts"
+topic_storebigquery="${survey}-BigQuery"
```
troyraen: Lower case when naming resources.

```suggestion
topic_storebigquery="${survey}-bigquery"
```

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+        logger.log_text(msg, severity="DEBUG")
+        table_dict = {"alerts_table": None}
+
+    return table_dict
+
+def _create_outgoing_alert(alert: pittgoogle.alert.Alert, table_dict: dict) -> pittgoogle.alert.Alert:
+    """Create an announcement of the table storage operation to Pub/Sub."""
+    # collect attributes
+    attrs = {
+        "alert_table": table_dict['alerts_table'],
```
troyraen: Call this `alerts_table` to be consistent with our other brokers.

```suggestion
        "alerts_table": table_dict['alerts_table'],
```

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+        "type": alert.dict['alert_type']
+    }
```
troyraen, line 84: Include the event ID. Otherwise there'll be no way to connect this outgoing alert with a row in the table.

```suggestion
        "superevent_id": alert.dict["superevent_id"],
    }
```

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+    decoded_data = base64.b64decode(event["data"])
+
+    # create a PubsubMessage-like object with the existing event dictionary
+    pubsub_message = pubsub_v1.types.PubsubMessage(
+        data=decoded_data,
+        attributes=event.get("attributes", {})
+    )
+
+    # unpack the alert
+    alert = pittgoogle.Alert.from_msg(msg=pubsub_message,schema_name="default_schema")
```
troyraen: Good for now. I'm in the process of adding a `from_cloud_functions` method that will take care of the decoding, etc. in pittgoogle-client.

broker/cloud_functions/lvk/store_BigQuery/main.py
```diff
@@ -0,0 +1,89 @@
...
+    """Insert rows into the `alerts` table via the streaming API."""
+    # send to bigquery
+    errors = TABLE.insert_rows([alert])
+
+    # handle errors; if none, save table id for Pub/Sub message
+    if len(errors) == 0:
+        table_dict = {"alerts_table": f"{TABLE.id}"}
+    else:
+        msg = f"Error inserting to alerts table: {errors}"
+        logger.log_text(msg, severity="DEBUG")
```
troyraen: I know this is whole block is just copied from the ZTF module but it's a really bad way to handle errors. Let's at least log this as a warning.

```suggestion
        logger.log_text(msg, severity="WARNING")
```

troyraen (changes requested):
A few things to clean up here, mostly around variable/resource naming and file formatting.

Please make these formatting changes on the files in this PR. I also recommend setting your IDE to do these things automatically every time you save a file. I think you use VSCode, so I'll list the settings for that.

- Run black on all python files. Use a max line length of 99.
    - Enable the setting "Editor: Format On Save".
    - Set black as the python formatter. I do this by setting "Editor: Default Formatter" to "Black Formatter". (You may need to install the black formatter.)
    - In "Black-formatter: Args", add the entry "--line-length=99".
    - If you want any of the above to apply only to this repo and not any others you work with in VSCode, set "Workspace" settings instead of "User" settings.
- Make all files end with exactly one blank line. Treat this as a universal standard. ([Why should text files end with a newline?](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline))
    - Enable settings "Files: Insert Final Newline" and "Files: Trim Final Newlines".
- Remove extra whitespace from the end of every line.
    - Enable the setting "Files: Trim Trailing Whitespace".

broker/cloud_functions/lvk/store_BigQuery/main.py
hernandezc1: The underscore was placed there to address a codacy issue (unused argument warning)

broker/setup_broker/lvk/setup_broker.sh
hernandezc1: GCN's [Schema Browser](https://gcn.nasa.gov/docs/schema/v4.1.0/gcn/notices#schema-browser) specifies the schema versions for GCN notices but it does not include an explicit schema version specifically for `lvk`. Since I couldn't find an "official" schema version I thought specifying the observation run associated with the schema would be more descriptive. 

I think the user should be specifying the schema associated with the observation run as you bring up: `schema_version="O4"`. What do you think @troyraen?

broker/cloud_functions/lvk/store_BigQuery/deploy.sh
```diff
@@ -0,0 +1,42 @@
+#! /bin/bash
+# Deploys or deletes broker Cloud Function
+# This script will not delete Cloud Functions that are in production
+
+testid="${1:-test}"
+# "False" uses production resources
+# any other string will be appended to the names of all resources
+teardown="${2:-False}"
+# "True" tearsdown/deletes resources, else setup
```
tregle, line 8: These comments seem improperly interleved with the variables they are documenting.

broker/cloud_functions/lvk/store_BigQuery/main.py
troyraen: I see, ok, that's a reasonable solution.

broker/setup_broker/lvk/setup_broker.sh
troyraen: I don't want to tie the observing run to the schema version if that's not what LVK has done. Yes that variable is and should be the schema version. If LVK really hasn't set one, then maybe we don't need this variable here?

broker/cloud_functions/lvk/store_BigQuery/deploy.sh
troyraen: They are, and this is a holdover of a mistake I made years ago.

broker/cloud_functions/lvk/store_BigQuery/deploy.sh
```diff
@@ -0,0 +1,42 @@
...
+survey="${3:-lvk}"
+# name of the survey this broker instance will ingest
+versiontag="${4:-vO4}"
```
troyraen:
```suggestion
# "False" uses production resources
# any other string will be appended to the names of all resources
testid="${1:-test}"
# "True" tearsdown/deletes resources, else setup
teardown="${2:-False}"
# name of the survey this broker instance will ingest
survey="${3:-lvk}"
versiontag="${4:-vO4}"
```

broker/setup_broker/lvk/setup_broker.sh
hernandezc1: Agreed. I've removed `schema_version` and `versiontag`

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -34,15 +34,31 @@ fi
...
+alerts_table="alerts"
+
+#--- Create (or delete) BigQuery, GCS, Pub/Sub resources
+echo
+echo "Configuring BigQuery, GCS, Pub/Sub resources..."
 if [ "${teardown}" != "True" ]; then
+    # create bigquery dataset and table
+    bq --location="${region}" mk --dataset "${bq_dataset}"
+
+    cd templates || exit
```
tregle: should have an exit code (and ideally, helpful message) to debug if an issue occurs.

broker/setup_broker/lvk/setup_broker.sh
```diff
@@ -34,15 +34,31 @@ fi
...
+    bq mk --table "${PROJECT_ID}:${bq_dataset}.${alerts_table}" "bq_${survey}_${alerts_table}_schema.json"
```
tregle: The better way to do this is to use a subshell. Then you're not changing the implicit global state.

```
cd templates
bq mk --table ....
```

-> `(cd templates && bq mk --table ....) || exit 5`

broker/cloud_functions/lvk/store_BigQuery/main.py
tregle, line 1: It would be good to have unit tests on this, but IIRC it's not a capability yinz have atm.

broker/setup_broker/lvk/templates/bq_lvk_alerts_schema.json
```diff
@@ -0,0 +1,216 @@
...
+          "type": "STRING"
+        },
+        {
+          "description": "time interval for which the signal is detected (above noise)",
+          "mode": "NULLABLE",
+          "name": "duration",
+          "type": "TIMESTAMP"
+        },
+        {
+          "description": "estimation of the frequency of the signal's main component",
```
tregle: units?

broker/setup_broker/lvk/templates/bq_lvk_alerts_schema.json
```diff
@@ -0,0 +1,216 @@
...
+          "mode": "NULLABLE",
+          "name": "central_frequency",
+          "type": "INTEGER"
+        },
+        {
+          "description": "Probability, under the assumption that the source is not noise, that at least one of the compact objects was a neutron star, that the system ejected a non-zero amount of neutron star matter, and that at least one of the compact objects has mass in the range 3-5 solar masses, respectively",
```
tregle: The description seems to just be a reiteration of the sub-columns.

It should probably describe the aggregate.

broker/setup_broker/lvk/templates/bq_lvk_alerts_schema.json
```diff
@@ -0,0 +1,216 @@
...
+            },
+            {
+              "description": "Probability that at least one of the compact objects has mass in the range 3-5 solar masses",
+              "mode": "REQUIRED",
+              "name": "HasMassGap",
+              "type": "FLOAT"
+            }
+          ],
+          "mode": "REQUIRED",
+          "name": "properties",
```
tregle: "properties" is pretty generic, though a good description could overcome that.

tregle (approved):
Improvements aren't really structural, so I'll all hit the approve bit.

Also, because I'm still new here. :)

broker/setup_broker/lvk/setup_broker.sh
troyraen: Thank you. We should make this replacement in a lot of places (#240).

broker/setup_broker/lvk/templates/bq_lvk_alerts_schema.json
troyraen, line 1: @hernandezc1 Add a comment at the top explaining where the information in this file came from and include a url. We would like to be able to point at a machine-readable schema definition (like an avro schema file) provided by the survey. But I think we haven't found one for LVK, is that right?

If you had to pull this together from info on their website, you probably had to make more decisions here than we usually do. @tregle pointed out good things to consider when creating a schema (thanks! It encouraged me to give this file a better look as well). Some of them we probably want to implement, but some of them we probably shouldn't because we aren't the data producers. (And a comment at the top can help make this clear.) For example, we don't want to change a field name, but we could add units to a description if it would make things more clear and we can find/point to them in LVK's docs. But you're the one with the most background here so you'll have to decide specifics.

One other question, How did you decide on the mode and type values here? If we have an avro schema, we can use it to create the bigquery table, and then download this json version for future use. That's the easiest way to make this schema match (as much as possible) the alerts' schema, which is really what it needs to do. (And we should have a bit of code in here that will do that. #239) If you haven't exhausted all the avenues yet, look/ask again to see if you can get your hands on a real schema definition from LVK. Avro would be great, but json or anything else they're using on the backend would be helpful (and a schema version?).

troyraen (approved):
I'd like to hear what the solution is for adding a table description with a link to the original schema. Otherwise, LGTM.

PR #236, opened 2024-07-22
Increase the max size of the `TempAlertFile` from 150k to 1M

troyraen:
Our Cloud Storage module uses a `SpooledTemporaryFile` to manipulate the alert before storing it in the bucket. ZTF alerts have started to exceed our original 150k limit, spilling over onto disk and raising warnings. This PR raises the limit to 1M.


wmwv (approved):
LGTM

PR #237, opened 2024-09-01
LVK Consumer: Pass Kafka metadata to Pub/Sub attributes

troyraen:
Our LVK consumer configuration is missing the setting that passes Kafka metadata like the topic, partition, offset, and timestamp through to the Pub/Sub message as metadata attributes. This PR adds the setting.


troyraen:
@hernandezc1 to test this, do we just have to upload the file to the LVK broker bucket and restart the consumer (like for ZTF)? Will you either test this or leave instructions on how to do it?

hernandezc1:
@troyraen Yes, that's exactly right. I've gone ahead and tested the changes:

Screenshot 2024-09-09 at 4 53 52 PM

PR #249, opened 2025-01-27
Use BigQuery subscriptions to store ZTF alert data

### LINKED ISSUE #244: Replace 'store_in_BigQuery' Cloud Function with a Pub/Sub to BigQuery Subscription
When the alert rate gets above about 20/sec we run into trouble with the Cloud Functions that send data to BigQuery. Replacing our main 'store_in_BigQuery' Cloud Function with a Pub/Sub [BigQuery Subscription](https://cloud.google.com/pubsub/docs/bigquery) is probably the easiest thing we can do that will (I hope) help significantly.

BigQuery itself should be able handle much larger rates; our trouble is a result of how we send the data (i.e., rate limits on certain API calls). The effect is that the execution time of a function to process a single alert increases because it must wait awhile before it can complete the API call. The problem increases with the alert rate. This can cause us to pay a lot more than we otherwise do to process those alerts.

Two other Cloud Functions ('tag' and 'SuperNNova') also send data to BigQuery and so contribute to the trouble, but let's start with the 'store_in_BigQuery' function because the solution is more straightforward. The same solution might work well for the others but we'll have to consider a couple more details.

hernandezc1:
This PR creates the necessary changes to write ZTF alert data and DIA source data directly to BigQuery using BigQuery subscriptions.

Prior to these changes, the [store_BigQuery cloud function](https://github.com/mwvgroup/Pitt-Google-Broker/blob/develop/broker/cloud_functions/store_BigQuery/main.py) transforms the original alert data published to the `ztf-alerts` topic (e.g., cutouts are dropped; DIA source data is extracted, etc.) and inserts rows into the `alerts` and `DIASource` table via the streaming API.

Here is an overview of the changes made in this PR:
- Update `setup_broker.sh`
     - creates new Pub/Sub topics and subscriptions to write data directly to BigQuery
     - removes redundant code; functions are now used to append `testid` (`define_GCP_resources`) if necessary and deploys/delete resources (`manage_resources`) based on the environment type (i.e., production or testing)
- Update `cloud_functions/store_BigQuery directory`
     - remove `schema_maps` directory (no longer needed)
     - update `deploy.sh`
     - update `requirements.txt` to include the pittgoogle-client
     - update `main.py` to transform the original alert data and publish the transformed data to the appropriate Pub/Sub topic; the attached BigQuery subscriptions will write the data directly to BigQuery

I think that names used to specify the resources created/changed in this PR could be more specific; can you provide feedback on this as well? (e.g., what should the store_BigQuery cloud function be renamed to?)

These changes have been successfully tested in the `ardent-cycling-243415` project. The trigger topic for the cloud function was `ztf-loop`.

Closes https://github.com/mwvgroup/Pitt-Google-Broker/issues/244


broker/setup_broker/setup_broker.sh
```diff
@@ -36,34 +36,101 @@ if [ "$continue_with_setup" != "y" ]; then
...
-fi
+broker_bucket=$(define_GCP_resources "${PROJECT_ID}-${survey}-broker_files")
+bq_dataset=$(define_GCP_resources "${survey}")
+# topics and subscriptions involved in writing DIASource data to BigQuery
+topic_diasource=$(define_GCP_resources "${survey}-diasource")
+subscription_diasource="${topic_diasource}" # BigQuery subscription
+topic_diasource_deadletter=$(define_GCP_resources "${survey}-diasource-deadletter")
+subscription_diasource_deadletter="${topic_diasource_deadletter}"
+# topics and subscriptions involved in writing alert data to BigQuery
+topic_alert_data=$(define_GCP_resources "${survey}-alert-data") # needs a better name
```
wmwv: I appreciate the point that `-data` doesn't seem like it's adding much.  Is this topic meant really to be used only by BigQuery?  If so, do you think it would be fine to call something specific like `ztf-alert-bigquery-import`?

wmwv (approved):
LGTM

I don't have strong feelings about the naming.  I appreciate your judgement that the name could be more useful.  I think my design advice would be to have specific, descriptive names and then some interface, visualization, list to make them easily discoverable.

hernandezc1:
@troyraen I've now removed any resources associated with the DIASource table. These changes have been tested using the `ztf-loop` topic and are ready for your review.

broker/setup_broker/setup_broker.sh
```diff
@@ -36,34 +36,80 @@ if [ "$continue_with_setup" != "y" ]; then
...
+            --dead-letter-topic-project="${PROJECT_ID}"
+    else
+        if [ "$environment_type" = "testing" ]; then
+            # delete testing resources
+            python3 setup_gcp.py --survey="$survey" --testid="$testid" --teardown --confirmed --versiontag="${versiontag}"
+            gcloud pubsub topics delete "${topic_bigquery_import}"
+            gcloud pubsub topics delete "${deadletter_topic_bigquery_import}"
+            gcloud pubsub subscriptions delete "${deadletter_subscription_bigquery_import}"
+            gcloud pubsub subscriptions delete "${subscription_bigquery_import}"
+        fi
```
troyraen: In case the user tries to delete production resources with this script, let's give them some info about why nothing will happen.

```suggestion
        if [ "$environment_type" = "testing" ]; then
            # delete testing resources
            python3 setup_gcp.py --survey="$survey" --testid="$testid" --teardown --confirmed --versiontag="${versiontag}"
            gcloud pubsub topics delete "${topic_bigquery_import}"
            gcloud pubsub topics delete "${deadletter_topic_bigquery_import}"
            gcloud pubsub subscriptions delete "${deadletter_subscription_bigquery_import}"
            gcloud pubsub subscriptions delete "${subscription_bigquery_import}"
        else
            echo 'ERROR: No testid supplied.'
            echo 'To avoid accidents, this script will not delete production resources.'
            echo 'If that is your intention, you must delete them manually.'
            echo 'Otherwise, please supply a testid.'
            exit 1
        fi
```

broker/setup_broker/setup_broker.sh
```diff
@@ -36,34 +36,80 @@ if [ "$continue_with_setup" != "y" ]; then
...
+manage_resources() {
+    local mode="$1"  # setup or teardown
+    local environment_type="production"
+
+    if [ "$testid" != "False" ]; then
+        environment_type="testing"
+    fi
 
+    if [ "$mode" = "setup" ]; then
+        # setup resources
```
troyraen, line 77: You're creating/deleting the new Pub/Sub resources directly in this script but all the others are handled in setup_gcp.py. That's fine (we probably want to migrate away from that python script), but add a comment stating that this is the case.

broker/setup_broker/setup_gcp.py
```diff
@@ -253,48 +256,49 @@ def setup_bigquery(survey='ztf', testid='test', teardown=False, versiontag="v3_3
...
                         # no data to load, just create the table
-                        bqmk = f'bq mk --table {table_id} {schema}.json'
+                        bqmk = f"bq mk --table {table_id} {schema}.json"
                         out = subprocess.check_output(shlex.split(bqmk))
-                        print(f'{out}')  # should be a success message
+                        print(f"{out}")  # should be a success message
+
 
-def setup_dashboard(survey='ztf', testid='test', teardown=False) -> None:
+def setup_dashboard(survey="ztf", testid="test", teardown=False) -> None:
```
troyraen: Go ahead and delete this whole function plus the related private functions and any other variables that would then be obsolete. I'm pretty sure it doesn't even work anymore but just throws an error. It would be great to have a dashboard, but this particular implementation was too fiddly to be sustainable. That should also take care of all the Codacy errors as well.

troyraen (approved):
Thanks! This is a major improvement because it avoids the BigQuery API calls that have caused us so much trouble when the incoming alert rate approaches that expected from LSST. I left some minor code comments below. Otherwise, LGTM.

I'm hopeful that we can drop the 'store_BigQuery' cloud fnc altogether and connect the BigQuery Subscription directly to the 'alerts' topic that is published by the bytes-to-bucket cloud fnc. Since that topic contains Avro alerts, I think (fingers crossed) BigQuery will do the right thing and just silently drop the cutouts instead of choking on them. Up to you whether you want to try that as part of this PR or not. Even without that, this is a great improvement for us, so either is fine with me.

troyraen:
Add the text "Closes #244" to the top of your main comment. That will link the issue to this PR and close it automatically when this is merged. If you leave the bigquery cloud fnc in with this PR you could open a separate issue to get rid of it (assuming the Avro -> BigQuery Subscription will work). But either way, #244 can be closed with this.

Issue #96, opened 2021-09-08
Potential issues at LSST-scale


troyraen:
### Cloud Functions

See [Quotas](https://cloud.google.com/functions/quotas), esp the section on background functions. 
Several have the potential to cause problems. 
Looking at one: "Max throughput limit"...

**TLDR: projecting ~45 minute delays, a few times each night**

- 10 MB per second: Max throughput of incoming events (Pub/Sub messages), per Cloud Function. This GCP quota cannot be increased.
- 80 KB per alert: projected size of LSST alert packets

This means:

- **125 alerts per second**: Max processing rate for any Cloud Function listening to our _alerts_ Pub/Sub stream. **This includes BigQuery and file storage components plus some of the science filters.**

But LSST's rate will easily surpass this:

- **278 alerts per second**: average LSST alert rate on the most active nights (10^7 alerts per night / 10 hours per night).

If ZTF is any indication... the **instantaneous** alert rate will often be higher than this. A quick glance at the recent [ZTF dashboard](https://monitor.alerts.ztf.uw.edu/d/XAs4s7Giz/ztf-alert-distribution-system-status?orgId=1&refresh=1m&from=now-7d&to=now) shows publish rates of >2x their equivalent active-night average (5x10^5 alerts / 10 hours) for 2-10 minute bursts a few times per night. 

Extrapolating to LSST:

- **~45 minutes**: length of time the last alerts in an equivalent 10 minute burst would wait before being processed by a single Cloud Function.  (2 x active-night average x 10 minutes / max processing rate)

wmwv:
Ask Ross about this technical question.  This must be solved in some way.

10 MB/s is a ridiculously low limit for lots of things that must be running on GCP; do none of those use Cloud Functions.

It's also possible this limit is more security policy minded, to avoid denial of resource problems, than it is a fundamental technical limitation.  There must be ways then to configuration the limits differently.

troyraen:
It might be that we should move these components to Cloud Run at that point. I have been spending a little time learning about that service recently.
Yes, I will ask Ross.

troyraen:
### Pipeline performance at LSST rates

On Sept 23, 2021 the incoming alert rate from ZTF spiked to near-LSST rates. I have documented the pipeline's performance [here](https://github.com/mwvgroup/Pitt-Google-Broker/blob/troy/troy/performance-figures/lsst-rates/README.md), including many figures. I've copied only the takeaways below:

[...]

PR #250, opened 2025-01-27
Create LSST storage pipeline (`ps_to_storage`)

hernandezc1:
This PR creates the necessary changes to store LSST alerts in Google Cloud Storage buckets using the `upload_bytes_to_bucket` Cloud Function.

Here is an overview of the changes made in this PR:
- Rename `setup_broker/rubin` directory to `setup_broker/lsst`
- Rename `consumer/rubin` directory to `consumer/lsst`
     - `PS_TOPIC_DEFAULT` for VM updated: `"${survey}-alerts` -> `"${survey}-alerts_raw"`
     - The `testid` is now appended to the `consumer.group.id` in the VM's configuration files to prevent new deployments from picking up existing offsets
- `setup_broker.sh` deploys all of the necessary GCP resources
- Create `cloud_functions/lsst/ps_to_gcs` directory
     - `main.py` deduplicates the "incoming" alert stream, stores the alert (.avsc) in a GCS bucket, and publishes the original alert bytes (schema not included) to the `lsst-alerts` Pub/Sub topic; the schema version of each alert is attached in the Pub/Sub message attributes. The BigQuery subscriptions in #243 can use the schema version as a filter to write the alert data to the appropriate BigQuery table


broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,196 @@
...
+VERSIONTAG = os.getenv("VERSIONTAG")
+
+# connect to the cloud logger and publisher
+logging_client = logging.Client()
+log_name = "ps-to-gcs-cloudfnc"
+logger = logging_client.logger(log_name)
+publisher = pubsub_v1.PublisherClient()
+
+# GCP resources used in this module
+client = storage.Client()
```
hernandezc1, line 34: Pre-defining the buckets and making a `.get_bucket` request for each _instance_ rather than each alert is a temporary solution I discussed with @troyraen. For future schema implementations, this will need to be updated

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,196 @@
...
+BUCKETS = {}
+
+# alerts are stored in GCS buckets based on their schema version; alerts in topic may contain multiple schema versions.
+# to avoid making the get_bucket call for each alert, we'll cache the buckets and assign the correct bucket dynamically
+# based on the alert's schema version
+for versiontag in available_schemas.values():
+    bucket_name = f"{PROJECT_ID}-{SURVEY}_alerts_{versiontag}"
+    if TESTID != "False":
+        bucket_name = f"{bucket_name}-{TESTID}"
+    BUCKETS[versiontag] = client.get_bucket(client.bucket(bucket_name, user_project=PROJECT_ID))
```
hernandezc1: Pre-defining the buckets and making a `.get_bucket` request for each _instance_ rather than each alert is a temporary solution I discussed with @troyraen. For future schema implementations, this will need to be updated

broker/cloud_functions/lsst/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,79 @@
...
+
+    #--- Create the bucket that will store the alerts
+    gsutil mb -l "${region}" "gs://${avro_bucket}"
+    gsutil uniformbucketlevelaccess set on "gs://${avro_bucket}"
+    gsutil requesterpays set on "gs://${avro_bucket}"
+    gcloud storage buckets add-iam-policy-binding "gs://${avro_bucket}" \
+        --member="allUsers" \
+        --role="roles/storage.objectViewer"
+
+    #--- Setup the Pub/Sub notifications on ZTF Avro storage bucket
```
troyraen:
```suggestion
    #--- Setup the Pub/Sub notifications on the Avro storage bucket
```

broker/cloud_functions/lsst/ps_to_gcs/deploy.sh
```diff
@@ -0,0 +1,79 @@
...
+        -e "$trigger_event" \
+        -f "$format" \
+        "gs://${avro_bucket}"
+    gcloud pubsub subscriptions create "${avro_subscription}" --topic="${avro_topic}"
+
+
+#--- Pub/Sub -> Cloud Storage Avro cloud function
+    echo "Deploying Cloud Function: ${ps_to_gcs_CF_name}"
+    ps_to_gcs_entry_point="run"
+    memory=512MB  # standard 256MB is too small here (it was always on the edge)
```
troyraen: Check to see if 512MB is really what this needs.

Regardless, let's get rid of this old comment of mine that keeps getting propagated but doesn't make sense out of the original context.

```suggestion
    memory=512MB  # standard 256MB is too small here
```

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,198 @@
...
+
+# define a binary data structure for packing and unpacking bytes
+_ConfluentWireFormatHeader = struct.Struct(">bi")
+
+# By default, spool data in memory to avoid IO unless data is too big
+# LSST alerts are anticipated at 80 kB, so 1000 kB should be plenty
+max_alert_packet_size = 1_000_000
```
troyraen:
```suggestion
```

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,198 @@
...
+    attributes = event.get("attributes", {})
+
+    # unpack the alert and read schema ID
+    header_bytes = alert_bytes[:5]
+    schema_id = deserialize_confluent_wire_header(header_bytes)
+
+    # get and load schema
+    sr_client = SchemaRegistryClient({"url": "https://usdf-alert-schemas-dev.slac.stanford.edu"})
+    schema = sr_client.get_schema(schema_id=schema_id)
+    latest_schema = json.loads(schema.schema_str)
```
troyraen: This will not necessarily be the latest schema.

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,198 @@
...
+BUCKETS = {}
+
+# alerts are stored in GCS buckets based on their schema version; alerts in topic may contain multiple schema versions.
+# to avoid making the get_bucket call for each alert, we'll cache the buckets and assign the correct bucket dynamically
+# based on the alert's schema version
+for versiontag in available_schemas.values():
+    bucket_name = f"{PROJECT_ID}-{SURVEY}_alerts_{versiontag}"
+    if TESTID != "False":
+        bucket_name = f"{bucket_name}-{TESTID}"
+    BUCKETS[versiontag] = client.get_bucket(client.bucket(bucket_name, user_project=PROJECT_ID))
```
troyraen:
```suggestion
# alerts are stored in GCS buckets based on their schema version; alerts in topic may contain multiple schema versions.
# to avoid making the get_bucket call for each alert, we'll cache the buckets and assign the correct bucket dynamically
# based on the alert's schema version
BUCKETS = {}
for versiontag in available_schemas.values():
    bucket_name = f"{PROJECT_ID}-{SURVEY}_alerts_{versiontag}"
    if TESTID != "False":
        bucket_name = f"{bucket_name}-{TESTID}"
    BUCKETS[versiontag] = client.get_bucket(client.bucket(bucket_name, user_project=PROJECT_ID))
```

Making it more clear that these two pieces of code are intimately related.

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,198 @@
...
+    # let it raise. the main function will catch it and then drop the message.
+    blob.upload_from_file(io.BytesIO(alert_bytes), if_generation_match=0)
+
+    # Cloud Storage says this is not a duplicate, so now we publish the broker's main "alerts" stream
+    return publish_outgoing_alert(
+        topic_name=ALERTS_TOPIC.name,
+        message=alert_bytes,
+        attributes={
+            "objectId": str(alert_dict["diaObject"]["diaObjectId"]),
+            "sourceId": str(alert_dict["diaSource"]["diaSourceId"]),
```
troyraen:
```suggestion
            "diaObjectId": str(alert_dict["diaObject"]["diaObjectId"]),
            "diaSourceId": str(alert_dict["diaSource"]["diaSourceId"]),
```

Important not to change the field names.

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def publish_outgoing_alert(
+    topic_name: str, message: bytes, attributes: Optional[dict] = None
+) -> str:
+    """Publish messages to a Pub/Sub topic."""
+
+    # enforce bytes type for message
+    if not isinstance(message, bytes):
+        raise TypeError("`message` must be bytes or a dict.")
```
troyraen: In this case (publishing our main `lsst-alerts` stream), we want this to be the original, Avro-serialized bytes. We don't want to publish a dict here.

broker/consumer/lsst/vm_startup.sh
```diff
@@ -57,14 +57,19 @@ fout_topics="${workingdir}/list.topics"
...
+fi
 
 cd "${workingdir}" || exit
 
 fconfig=admin.properties
 sed -i "s/KAFKA_PASSWORD/${KAFKA_PASSWORD}/g" ${fconfig}
 
 fconfig=psconnect-worker.properties
-sed -i "s/KAFKA_PASSWORD/${KAFKA_PASSWORD}/g" ${fconfig}
+sed -i "s/KAFKA_PASSWORD/${KAFKA_PASSWORD}/g" ${fconfig} && sed -i "s/GROUP_ID/${group_id}/g" ${fconfig}
```
troyraen:
```suggestion
sed -i "s/KAFKA_PASSWORD/${KAFKA_PASSWORD}/g" ${fconfig}
sed -i "s/GROUP_ID/${group_id}/g" ${fconfig}
```

No reason to make the latter dependent on the former.

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -0,0 +1,146 @@
...
+        # create a firewall rule to open the port used by Kafka/Rubin LSST
+        # on any instance with the flag --tags=tcpport9094
+        echo
+        echo "Configuring Rubin/Kafka firewall rule..."
+        firewallrule="tcpport9094"
+        gcloud compute firewall-rules create "${firewallrule}" \
+            --allow=tcp:9094 \
+            --description="Allow incoming traffic on TCP port 9094" \
+            --direction=INGRESS \
+            --enable-logging
```
troyraen, line 103: Is this necessary for LSST?

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -0,0 +1,146 @@
...
+
+    echo "${base_name}${testid_suffix}"
+}
+
+#--- GCP resources used directly in this script
+broker_bucket=$(define_GCP_resources "${PROJECT_ID}-${survey}-broker_files")
+topic_alerts_raw=$(define_GCP_resources "${survey}-alerts_raw")
+subscription_alerts_raw=$(define_GCP_resources "${survey}-alerts_raw-counter")
+topic_alerts=$(define_GCP_resources "${survey}-alerts")
+subscription_alerts=$(define_GCP_resources "${survey}-alerts-counter")
```
troyraen: Consider whether you actually want/need these two "counter" subscriptions. Do you plan to use them for something? If not, better to avoid the cost by just not creating them since these will be very large streams.

broker/setup_broker/lsst/upload_broker_bucket.sh
troyraen, line 1: This script is uploading the entire 'consumer' directory. I think it would be better to upload only the 'consumer/lsst' directory.

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+
+"""This module stores the alert data as an Avro file in Cloud Storage,
+fixing the schema first, if necessary.
+"""
```
troyraen:
```suggestion
"""This module stores the alert data as an Avro file in Cloud Storage."""
```

troyraen (commented):
Great! Mostly minor suggestions, but pay special attention that field names are not being changed and to the Cloud Function's memory allotment.

Also, I've been thinking that it would be better to have just one 'alerts' bucket per survey. That would be more user friendly and also a bit easier for us to maintain. We definitely want to be able to easily grab all the alerts for a given schema version (e.g., to regenerate a BigQuery table), but we could make that possible by making a top-level "directory" for each schema version. The user would still have to know what the schema version is in order to construct the file path, but at least they wouldn't have to connect to a totally different bucket. (pittgoogle-client should hide all of that so its users don't need to know either the bucket name or the schema version, but not all Pitt-Google users will use pittgoogle-client.)

If we're going to do that, we should do it for LSST now before we starting putting alerts in the bucket. @hernandezc1 @wmwv what do you think?

broker/cloud_functions/lsst/ps_to_gcs/deploy.sh
hernandezc1: With the tests I've ran this week, this cloud function uses ~330MB of memory at its peak

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: I don't have a strong argument to keep them. Will remove

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: I'm not entirely sure. The firewall rule allows incoming traffic on TCP port 9094. If this code block is removed, I imagine that incoming traffic is now prevented from reaching the port(?)

hernandezc1:
@troyraen Thanks for your feedback! I've gone ahead and implemented these changes. Notably, a single 'alerts' bucket is now created and the `upload_bytes_to_bucket` Cloud Function specifies a top-level "directory" for each schema version when generating a filename.

broker/setup_broker/lsst/setup_broker.sh
troyraen: Sorry, I think I was conflating the firewall and the authentication mechanisms in my head when I wrote this.

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    Generate the filename of an alert stored to a Cloud Storage bucket.
+
+    Args:
+        aname:
+            Components to create the filename. Required key/value pairs are those needed to create a parsed filename.
+            Extra keys are ignored.
+
+    Returns:
+        str: The formatted filename as "{topic}/{objectId}/{sourceId}.{format}".
```
troyraen: Update the docstring.

broker/cloud_functions/lsst/ps_to_gcs/main.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    schema_version = aname.get("schema_version")
+    topic = aname.get("topic", "no_topic")
+    object_id = aname.get("objectId")
+    source_id = aname.get("sourceId")
+    file_format = aname.get("format", "avro")
```
troyraen:
```suggestion
    schema_version = aname["schema_version"]
    alert_date = aname["alert_date"]
    object_id = aname["objectId"]
    source_id = aname["sourceId"]
    file_format = aname["format"]
```

Using `get` here is probably a holdover from the ZTF fnc but, looking at it now, I don't like it because it can hide bugs that may crop up as this code gets copy/pasted for new surveys. I think it would be better to let this function throw a `KeyError` if it doesn't get the info that is required to construct the object name.

For the "topic"... That level in the file name should be the date (UTC) that the alert was published. For ZTF, the topic name was convenient because it includes the date, but it was a bad choice because it made this code brittle... Since LSST isn't naming their topics that way, this won't work here the way we want it to. I think we need to get the date from the appropriate field in the alert packet. I recommend that we format it as YYYY-MM-DD so that it's easy to search for all alerts from a given year or year+month. As I recall, a wildcard like `*` can only be used at the end of the string when searching GCS. Even if it accepts wildcards in other places, my guess is that those searches are much slower.

troyraen (approved):
Thanks! I noticed a couple more things about the way we're generating the object names that we should change, especially the "topic". From experience, it is a giant pain to deal with millions of objects in a bucket when they're not well organized (i.e., named), so worth it to spend time on that up front.

Otherwise, LGTM.

troyraen:
Oh... Since this is for LSST, I think it will need to be a Cloud Run service rather than a Function. See #96. You can try it as-is if you want, but watch it closely. I think it will run into significant problems due to rate issues.

hernandezc1:
@wmwv @troyraen new changes have been implemented and are ready for review. This PR now uses Cloud Run instead of Cloud Functions and incorporates the changes in #243 (PR will be closed).

wmwv (approved):
LGTM

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -1,37 +1,53 @@
...
+
+# Variables for outgoing data
+HTTP_204 = 204  # HTTP code: Success
+HTTP_400 = 400  # HTTP code: Bad Request
 
 # GCP resources used in this module
 ALERTS_TOPIC = pittgoogle.Topic.from_cloud(
     "alerts", survey=SURVEY, testid=TESTID, projectid=PROJECT_ID
 )
+TOPIC_BIGQUERY_IMPORT = pittgoogle.Topic.from_cloud(
```
troyraen:
```suggestion
ALERTS_TOPIC = pittgoogle.Topic.from_cloud(
    "alerts", survey=SURVEY, testid=TESTID, projectid=PROJECT_ID
)
BIGQUERY_IMPORT_TOPIC = pittgoogle.Topic.from_cloud(
```

I recommend keeping the syntax the same for clarity. (Alternately, could switch `ALERTS_TOPIC` -> `TOPIC_ALERTS`, which might actually be better both because it would match the `ROUTE_RUN` syntax and because it puts the parts in order of increasing specificity.)

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -42,38 +58,44 @@
...
+    # this contains a single Pub/Sub message with the alert to be processed
+    envelope = flask.request.get_json()
+
     try:
-        upload_bytes_to_bucket(event, context)
+        store_alert_data(envelope)
     # this is raised by blob.upload_from_file if the object already exists in the bucket
     except PreconditionFailed:
         # we'll simply pass, and the duplicate alert will go no further in our pipeline
         pass
```
troyraen: I know that past me wrote it like this, but current me sees way too many things happening between `try` and `except`. Ideally, only the thing that might throw the error (i.e., uploading to the bucket) should be in the `try` block. Just FYI.

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -93,20 +115,20 @@ def upload_bytes_to_bucket(event: dict, context: functions_v1.context.Context) -
             "schema_version": schema_version,
```
troyraen: What does `schema_version` look like? Since it will be used to name the alert in the bucket, hopefully it's clear from the value alone (a) that it is identifying a schema version, and (b) which specific schema version it maps to if one is looking in LSST's docs/repos. Just pointing it out as a thing that I think is important to consider before writing a gazzilion alerts to the bucket. It's probably fine, I just can't tell based on the code alone. (In particular, seeing it generated with `.split(".")[1]` is unexpected to me and makes me wonder what the result looks like.)

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -146,31 +173,38 @@ def generate_alert_filename(aname: dict) -> str:
...
-    return f"{schema_version}/{topic}/{object_id}/{source_id}.{file_format}"
+    schema_version = aname["schema_version"]
+    alert_date = aname["alert_date"]
+    object_id = aname["objectId"]
+    source_id = aname["sourceId"]
+    file_format = aname["format"]
 
+    # convert the MJD timestamp to "YYYY-MM-DD"
+    time_obj = Time(alert_date, format="mjd")
+    date_string = time_obj.datetime.strftime("%Y-%m-%d")
```
troyraen: Rather than splitting the work between `store_alert_data` and this function, I would either do this conversion in `store_alert_data` so that this function remains generically usable for any survey or else just pass in the whole `alert_dict` and let this function pull out and convert everything it needs. Functionally, it makes no difference here, but it would help in the future if/when this code gets reused.

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -93,20 +115,20 @@ def upload_bytes_to_bucket(event: dict, context: functions_v1.context.Context) -
...
         }
     )
 
     blob = bucket.blob(filename)
-    blob.metadata = create_file_metadata(alert_dict, context)
+    blob.metadata = create_file_metadata(alert_dict, event_id=envelope["message"]["messageId"])
 
     # raise a PreconditionFailed exception if filename already exists in the bucket using "if_generation_match=0"
     # let it raise. the main function will catch it and then drop the message.
     blob.upload_from_file(io.BytesIO(alert_bytes), if_generation_match=0)
```
troyraen: Using `upload_from_file` is a holdover from when this function used a temporary file. Since that's gone now, it would be more direct to use [upload_from_string](https://cloud.google.com/python/docs/reference/storage/1.37.1/blobs#uploadfromstringdata-contenttypetextplain-numretriesnone-clientnone-predefinedaclnone-ifgenerationmatchnone-ifgenerationnotmatchnone-ifmetagenerationmatchnone-ifmetagenerationnotmatchnone-timeout60-checksumnone) and drop `io.BytesIO`.

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -1,37 +1,53 @@
...
+from google.cloud import logging, storage, pubsub_v1
 from google.cloud.exceptions import PreconditionFailed
 
+# [FIXME] Make this helpful or else delete it.
+# Connect the python logger to the google cloud logger.
+# By default, this captures INFO level and above.
+# pittgoogle uses the python logger.
+# We don't currently use the python logger directly in this script, but we could.
+logging.Client().setup_logging()
+publisher = pubsub_v1.PublisherClient()
```
troyraen: Use either the `pittgoogle.Topic`s or this, not both. (Preferably, `pittgoogle.Topic`. And if it's troublesome to use `pittgoogle.Topic.publish` since you're not using a `pittgoogle.Alert`, you can still avoid this by getting the client from `pittgoogle.Topic.client`.)

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -1,37 +1,53 @@
...
 
 PROJECT_ID = os.getenv("GCP_PROJECT")
 TESTID = os.getenv("TESTID")
 SURVEY = os.getenv("SURVEY")
 VERSIONTAG = os.getenv("VERSIONTAG")
```
troyraen: This is not being used in this module. (Does it even make sense to use this for LSST modules in general? IDK. Something to consider.)

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -42,38 +58,44 @@
...
 
-def run(event: dict, context: functions_v1.context.Context) -> None:
-    """Entry point for the Cloud Function
 
-    For args descriptions, see:
-    https://cloud.google.com/functions/docs/writing/background#function_parameters
+@app.route(ROUTE_RUN, methods=["POST"])
+def run():
+    """Uploads alert data to a GCS bucket. Publishes a de-duplicated "alerts" stream and publishes valid JSON message
+    to a Pub/Sub topic.
```
troyraen: It would be nice to be more clear about the "publishes valid JSON message to a Pub/Sub topic" bit.

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -180,3 +214,35 @@ def publish_outgoing_alert(
     future = publisher.publish(topic_path, data=message, **attributes)
 
     return future.result()
+
+
+def _create_valid_json(alert_dict: dict, attributes: dict) -> pittgoogle.alert.Alert:
```
troyraen: The name of this function makes me think that it returns a json string, but it doesn't.

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -180,3 +214,35 @@ def publish_outgoing_alert(
...
+    cutouts = [
+        "cutoutTemplate",
+        "cutoutScience",
+        "cutoutDifference",
+    ]
+    for key in cutouts:
+        alert_dict.pop(key, None)
+
+    # replace NaN values with None
+    valid_json = _transform_nan_to_none(alert_dict)
```
troyraen: The name of this variable makes me think that it is a json string, but it isn't.

broker/cloud_functions/lsst/ps_to_storage/main.py
```diff
@@ -180,3 +214,35 @@ def publish_outgoing_alert(
...
+
+    return pittgoogle.Alert.from_dict(payload=valid_json, attributes=attributes)
+
+
+def _transform_nan_to_none(alert_dict: dict) -> dict:
+    """Recursively replace NaN values with None in a dictionary."""
```
troyraen: The argument name, the type hint, and the docstring all indicate that the argument must be a dict, but that's not how it's used inside the function. 

The return type indicates that the returned value will always be a dict, but it's not.

This might be easier to make clear by separating into two functions: one that is intended to be called directly (with `alert_dict`) and another that handles the various types found inside the dict.

broker/consumer/lsst/vm_install.sh
```diff
@@ -68,8 +68,8 @@ wget https://github.com/GoogleCloudPlatform/pubsub/releases/download/${CONNECTOR
 echo "Done installing the Kafka -> Pub/Sub connector"
 
 #--- Set the startup script and shutdown
-startupscript="gs://${broker_bucket}/consumer/${survey}/vm_startup.sh"
+startupscript="gs://${broker_bucket}/${survey}/vm_startup.sh"
```
troyraen, line 71: What's going on here and in the several places below where the 'consumer' level is being removed?

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -88,19 +105,59 @@ manage_resources() {
...
         gcloud pubsub topics create "${topic_alerts_raw}"
         gcloud pubsub topics create "${topic_alerts}"
+        gcloud pubsub topics create "${topic_bigquery_import}"
+        gcloud pubsub topics create "${deadletter_topic_bigquery_import}"
         gcloud pubsub subscriptions create "${subscription_reservoir}" --topic="${topic_alerts}"
-
+        gcloud pubsub subscriptions create "${deadletter_subscription_bigquery_import}" --topic="${deadletter_topic_bigquery_import}"
+        # in order to create BigQuery subscriptions, ensure that the following service account:
+        # service-<project number>@gcp-sa-pubsub.iam.gserviceaccount.com" has the
+        # bigquery.dataEditor role for each table
```
troyraen, line 115: I'm confused by this comment. The way it's worded makes me think that this is what the next line does, but it doesn't. Looking a few lines down, I see that this service account is used explicitly but not in a way that seems to be implementing these instructions. Is the intention that user does this manually before running this script? If so, can that code just be added here?

broker/setup_broker/lsst/setup_broker.sh
troyraen, line 1: At some point, we really need to separate this script into three or more. Some things in here are one-time (per survey) broker setup tasks, which was the original intention for this script. Other things need to be run every time there is a new schema -- It would be nice to have a dedicated script for that. Still other things need to be run once per GCP Project -- I think we have a script started for this in the user-demos repo which we could copy over here and expand. Mentioning it for you to consider next time you need to make mass updates to this script or need to make this script for a new survey.

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -88,19 +105,59 @@ manage_resources() {
...
+            # this allows dead-lettered messages to be forwarded from the BigQuery subscription to the dead letter topic
+            # and it allows dead-lettered messages to be published to the dead letter topic.
+            PUBSUB_SERVICE_ACCOUNT="service-${PROJECT_NUMBER}@gcp-sa-pubsub.iam.gserviceaccount.com"
+            gcloud pubsub topics add-iam-policy-binding "${deadletter_topic_bigquery_import}" \
+                --member="serviceAccount:$PUBSUB_SERVICE_ACCOUNT"\
+                --role="roles/pubsub.publisher"
+            gcloud pubsub subscriptions add-iam-policy-binding "${subscription_bigquery_import}" \
+                --member="serviceAccount:$PUBSUB_SERVICE_ACCOUNT"\
+                --role="roles/pubsub.subscriber"
+        fi
```
troyraen: Aren't these policy bindings also need in non-production deployments? It seems to me that it's actually the policy binding being done right above this `if` statement that should be inside the `if` statement, and these policy bindings should not be.

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -88,19 +105,59 @@ manage_resources() {
...
+
+        #--- Create Artifact Registry Repository
+        echo
+        echo "Configuring Artifact Registry..."
+        gcloud artifacts repositories create "${artifact_registry_repo}" --repository-format=docker \
+            --location="${region}" --description="Docker repository for Cloud Run services" \
+            --project="${PROJECT_ID}"
+        gcloud auth configure-docker "${region}"-docker.pkg.dev # authenticate requests to Artifact Registry
```
troyraen: What does this do? Reading the docs page, it seems like it either makes/changes a Docker config file that is specific to the user (i.e., is on their local machine and not in this repo) or makes/changes a Docker file that is specific to the Cloud Run module. If the former, this line probably belongs here: https://pitt-broker.readthedocs.io/en/latest/broker/run-a-broker-instance/initial-setup.html#setup-local-environment. If the later, can we just put the required change into module where its effect will be more clear?

broker/cloud_functions/lsst/ps_to_storage/main.py
troyraen, line 1: Line 105 creates a `SchemaRegistryClient`. It looks like something that makes an external call and so would be better to do once at the top of the script rather than inside the function where it will execute for every alert.

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -52,10 +52,19 @@ define_GCP_resources() {
 }
 
 #--- GCP resources used directly in this script
+artifact_registry_repo=$(define_GCP_resources "cloud-run-services")
```
troyraen: If this repo is intended for use by our LSST broker only, then "lsst" should appear in its name. If this is really a Project setup task and the repo is intended for use by all of our brokers that operate in the same region, then this script should not delete it in the "teardown" section below.

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -88,19 +105,59 @@ manage_resources() {
...
             gsutil -m -o "${o}" rm -r "gs://${broker_bucket}"
+            bq rm -r -f "${PROJECT_ID}:${bq_dataset}"
             gcloud pubsub topics delete "${topic_alerts_raw}"
             gcloud pubsub topics delete "${topic_alerts}"
+            gcloud pubsub topics delete "${topic_bigquery_import}"
+            gcloud pubsub topics delete "${deadletter_topic_bigquery_import}"
             gcloud pubsub subscriptions delete "${subscription_reservoir}"
+            gcloud pubsub subscriptions delete "${deadletter_subscription_bigquery_import}"
+            gcloud pubsub subscriptions delete "${subscription_bigquery_import}"
+            gcloud artifacts repositories delete "${artifact_registry_repo}" --location="${region}"
```
troyraen, line 165: Does this belong here? See my comment above.

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -125,15 +182,15 @@ echo
...
 echo "Configuring Cloud Functions..."
 cd .. && cd .. || exit
 cd cloud_functions && cd lsst || exit
 
-#--- Pub/Sub -> Cloud Storage Avro cloud function
-cd ps_to_gcs || exit
-./deploy.sh "$testid" "$teardown" "$survey" "$versiontag" "$region"
+#--- ps_to_storage cloud function
+cd ps_to_storage || exit
+./deploy.sh "$testid" "$teardown" "$survey" "$region"
```
troyraen: Seems like "Functions" and "function" don't belong here.

broker/cloud_functions/lsst/ps_to_storage/deploy.sh
```diff
@@ -9,11 +9,12 @@ testid="${1:-test}"
 teardown="${2:-False}"
 # name of the survey this broker instance will ingest
 survey="${3:-lsst}"
-# schema version
-versiontag="${4:-v7_3}"
-region="${5:-us-central1}"
+region="${4:-us-central1}"
 PROJECT_ID=$GOOGLE_CLOUD_PROJECT # get the environment variable
 
+MODULE_NAME="to-storage"  # lower case required by cloud run
```
troyraen:
```suggestion
MODULE_NAME="alerts_to_storage"  # lower case required by cloud run
```

Our standard syntax is to use an underscore to separate the words in the base name of a module. The survey name then gets prepended with a dash. That way, it is easy to write code that parses them. I also suggest to add "alerts" or something similar to the name, since this is not the only module that will send data to storage.

troyraen (changes requested):
Thanks! I have quite a few suggestions/requests in code comments below -- here is a list of the ones that I think are most crucial and should be addressed before merging:

- Make sure the value of the schema version that is used to name the alert in the bucket is clear.
- Use our standard syntax for the module's base name.
- Create the `SchemaRegistryClient` at the top of the script, assuming it really is making an external call.
- Either the name or the handling of the artifact registry repo needs to change.
- `gcloud auth configure-docker` seems like it doesn't belong in the setup script.

Plus two that are not in code comments below:

- The name of the 'cloud_functions' directory needs to be changed to 'cloud_run'.
- Reminder: Check the Avro files that end up in the bucket -- make sure that you can open them as expected.

And a final comment, which I'll link to from your pittgoogle-client PR in case you want to implement it there:

There is quite a bit of parsing and loading in here that could be avoided by using a `pittgoogle.Alert`. Too bad that the LSST schemas are a little tricky and we don't quite have them implemented in pittgoogle-client yet in a way that will work smoothly here. I just took a quick look at pittgoogle-client and I think what we should do is alter `lsst_auto_schema_helper` to remove the dependency on `lsst.alert.packet.schema` (since it isn't available and seems like it probably won't be anytime soon) and use the functionality of our `lsst_schema_helper` in its place. In other words, merge `lsst_auto_schema_helper` and `lsst_schema_helper`. Then the user (in this case, this module) will not have to know the schema version and can just do `alert = pittgoogle.Alert.from_cloud_run(envelope, schema_name='lsst')`.

broker/cloud_functions/lsst/ps_to_storage/main.py
hernandezc1: The output of `schema_version` is of the form `v7_x` where x is the minor version (e.g., `v7_3`). More specifically, parse_schema["namespace"] (in line 108) has the value `lsst.v7_x`

broker/consumer/lsst/vm_install.sh
hernandezc1: > This script is uploading the entire 'consumer' directory. I think it would be better to upload only the 'consumer/lsst' directory.

Related to the [comment above](https://github.com/mwvgroup/Pitt-Google-Broker/pull/250#discussion_r1962324559). Line 8 in `upload_broker_bucket.sh` was updated: 
`gsutil -m -o "${o}" cp -r ../../consumer "gs://${broker_bucket}"` -> `gsutil -m -o "${o}" cp -r ../../consumer/lsst "gs://${broker_bucket}"`

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: The repo should be deleted if it is a testing resource (i.e., it has a `testid` value). Will update to include the survey name

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: In order to deploy our Cloud Run service, we first need to:
- [create a Docker Repository](https://cloud.google.com/artifact-registry/docs/docker/store-docker-container-images#create) to store the container image (lines 143-145)
- configure Docker to use the Google Cloud CLI to [authenticate requests](https://cloud.google.com/artifact-registry/docs/docker/store-docker-container-images#auth) to Artifact Registry (line 146). This updates our Docker configuration and allows us to connect with Artifact Registry to push and pull images to/from the repo in our GCP project

Since our broker will deploy multiple container images (e.g., one needed for this Cloud Run service and another for SuperNNova) I think it would be better suited to leave this here. Otherwise, the deployment scripts for each service will have to create their own Artifact Registry repo (this is technically fine, but may become more confusing as the number of repos--and images--increase.

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: Agreed, I think there's a lot of room for improvement. My goal is to switch to using Terraform altogether after setting up our storage & science pipelines for LSST

broker/cloud_functions/lsst/ps_to_storage/main.py
hernandezc1: It's needed to set up the BigQuery tables, but not needed explicitly here. Thanks for catching

broker/consumer/lsst/vm_install.sh
troyraen: Ah, my confusion came from our file organization. I was looking at broker/setup_broker/upload_broker_bucket.sh, but LSST has it's own version of that script in a different directory. Would be nice to make ZTF directories to match the LSST and LVK directories so that it's clear that those files are only used for ZTF. (So much to do, so little time.)

broker/setup_broker/lsst/setup_broker.sh
troyraen: My specific question here is about line 146. (GitHub is providing context by also showing the three previous lines but I actually only selected 146 when commenting.) In my experience, `gcloud auth` configures the user's authentication settings. If that is what's happening here, then I think this line belongs in https://pitt-broker.readthedocs.io/en/latest/broker/run-a-broker-instance/initial-setup.html#setup-local-environment.

broker/cloud_functions/lsst/ps_to_storage/deploy.sh
hernandezc1: "alerts" added to the name! 

Underscores are not permitted when naming Cloud Run services. I am having trouble finding where it is explicitly stated in Google's documentation, but I can confirm this is true through personal experience. The following error appears when underscores are used: 
`The name must use only lowercase alphanumeric characters and dashes, cannot begin or end with a dash, and cannot be longer than 63 characters.`

broker/setup_broker/lsst/setup_broker.sh
troyraen: > The repo should be deleted if it is a testing resource

Hmm, I think that still has the potential to cause problems since 2+ different broker testing instances can be running in the same project simultaneously. Perhaps we should treat this like the firewall where this script creates it but does not delete it. It will only actually be created once in a given project but since we don't have a different script for that here right now, this script can just go ahead and try to create it every time.

broker/cloud_functions/lsst/ps_to_storage/deploy.sh
troyraen: Ah, ok. Nothing we can do about that then.

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: Providing this for reference (these may be the docs you read already):
- more information on the [gcloud auth configure-docker command](https://cloud.google.com/sdk/gcloud/reference/auth/configure-docker)
- more info on the [gcloud CLI credential helper](https://cloud.google.com/artifact-registry/docs/docker/authentication#gcloud-helper), which involves the command in line 146

From my interpretation of what I'm seeing, I don't believe line 146 is configuring a user's authentication settings in gcloud itself. Rather, it configures Docker to authenticate to Artifact Registry hosts in any environment where the [Google Cloud CLI](https://cloud.google.com/sdk/docs) is installed. In this case, `"${region}"-docker.pkg.dev`

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: Can you elaborate on the issue? 

If there are multiple testing instances in the same project, then the respective Artifact Registry repos should all have unique names due to the appended `testid`. I.e., the images in the Artifact Registry repo for different testing broker instances in the same project will be independent of one another. Deleting one repo, or manipulating its content, will not affect the content of a separate Artifact Registry repo in the same project.

broker/cloud_functions/lsst/ps_to_storage/main.py
hernandezc1: What I currently have set up is the only thing that works at the moment. If I get the client from `pittgoogle.Topic.client`, I need to specify the topic name (`pittgoogle.Topic(TOPIC_ALERTS).client`) since `publisher = pittgoogle.Topic.client` is a property and not an instance of `google.cloud.pubsub_v1.PublisherClient`. This method results in an error: "PermissionError: Cannot obtain credentials." and requires me to pass `GOOGLE_APPLICATION_CREDENTIALS` as an environment variable. Since this issue should be addressed when updating the pittgoogle-client, I'd like to keep things as they are. If you have any other suggestions I'd be happy to test them out

hernandezc1:
> * Either the name or the handling of the artifact registry repo needs to change.
> * `gcloud auth configure-docker` seems like it doesn't belong in the setup script.

@troyraen I've implemented new changes which are ready for your review. I think there's still more to discuss with the two items listed above

broker/setup_broker/lsst/setup_broker.sh
hernandezc1: That was my original intention but I've now added the code here!

broker/setup_broker/lsst/setup_broker.sh
troyraen: Oh sorry, yes a testid will keep them separate.

broker/cloud_functions/lsst/ps_to_storage/main.py
troyraen: You're already creating instances of `pittgoogle.Topic` here, both `TOPIC_ALERTS` and `TOPIC_BIGQUERY_IMPORT`. The latter is actually being used to publish a message but `TOPIC_ALERTS` is only being used to get the name of that topic back out again. My suggestion is to also use `TOPIC_ALERTS.client` rather than creating this separate client.

broker/setup_broker/lsst/setup_broker.sh
troyraen: From that first link:
> gcloud auth configure-docker adds the Docker credHelper entry to Docker's configuration file, or creates the file if it doesn't exist.

Where does Docker's configuration file live? What is its scope? My first guess was that it lives on the user's machine, probably someplace like `$HOME/.docker`, and is thus scoped to the user. If it does live in the GCP Project somewhere, is it scoped to the user, the Project, the module, or ...?

troyraen (approved):
Thanks for all this good work. Take another look at these two things, otherwise LGTM.

- The pubsub publisher client. Should be an easy change to use `TOPIC_ALERTS.client`.
- `gcloud auth configure-docker` -- If the effects of this are scoped to the user then this command should go in the user setup docs. It's probably not crucial enough that it needs to hold up this PR too much longer, but take one more look at it. The broker setup script is already doing too many different kinds of things and I'd like to avoid mixing in user setup tasks as well.

PR #251, opened 2025-02-22
Add SuperNNova to LSST science pipeline

hernandezc1:
This PR introduces a new module, _SuperNNova_, to the broker's LSST science pipeline. It incorporates all of the relevant services (i.e., classification and storage of classification results) currently in production for ZTF using the `pittgoogle-client` and Cloud Run (instead of Cloud Run Functions, previously known as Cloud Functions).

Here is an overview of this module:
- The `lsst-lite` stream triggers a Cloud Run service (`lsst-supernnova`) that performs a binary classification (Ia vs. non-Ia)
- Outgoing alerts are published to: `lsst-SuperNNova`:
     - Messages published to this topic are organized using two top level keys: `alert_lite` and`SuperNNova`.
     - The `alert_lite` value contains the original fields of the lite alert packet; the value of the `SuperNNova` key contains the `diaObjectId`, `diaSourceId`, `prob_class0`, `prob_class1`, and `predicted_class`.
- A BigQuery subscription attached to the `lsst-SuperNNova` topic writes the `SuperNNova` value to a BigQuery table: SuperNNova (dataset: lsst_value_added).


hernandezc1:
@troyraen requesting your review. The code in `main.py` has been tested on my local device, but I am having issues with the Cloud Run service due to dependency conflicts (the Codacy issues are associated with this). There's still work that needs to be done, but thought it would be worthwhile to have you check the progress.

troyraen:
> having issues with the Cloud Run service due to dependency conflicts (the Codacy issues are associated with this)

Let me know what you find there. I just disabled the two codacy patterns related to pinning versions before seeing your message. (I recently went thru and changed the tools and patterns we use in codacy and so I figured these were just popping up as new issues because of a new tool I chose.) In general I think I'd rather leave them unpinned but we may want to pin particular one(s) that have dependency conflicts.

broker/cloud_run/lsst/classify_snn/main.py
```diff
@@ -0,0 +1,170 @@
...
+    # classify
+    snn_dict = _classify(alert)
+
+    # prepare data for publishing
+    classifier_summary = _classification_summary(snn_dict)
+    snn_alert = _create_outgoing_alert(alert, snn_dict)
+    snn_results = pittgoogle.Alert.from_dict(payload=snn_dict)
+
+    # publish
+    TOPIC.publish(snn_alert, serializer="avro")
```
troyraen: Does this actually work with the avro serializer? I think it's going to try to use LSST's schema and that won't match snn_alert.

broker/cloud_run/lsst/classify_snn/main.py
```diff
@@ -0,0 +1,170 @@
...
+    return snn_df
+
+
+def _create_outgoing_alert(alert: pittgoogle.Alert, snn_dict: dict) -> pittgoogle.Alert:
+    """Combine the incoming alert with the classification results to create the outgoing alert."""
+
+    return pittgoogle.Alert.from_dict(
+        payload={**alert.dict, "SuperNNova": snn_dict},
+        attributes={"supernnova_class": snn_dict["predicted_class"], **alert.attributes},
+    )
```
troyraen: If we're going with a "lite" structure that looks like this:

```
'alert': {...}
'value-added': {our stuff...}
```

then let's do the same thing here. But tbh, I'm not totally sold on that because I think it's going to complicate a few things for us. One is how the `Alert` methods handle the data. All of a sudden there's this extra layer that it's not expecting.

Either way, I think `pittgoogle.Alert` should have a method that accepts a value-added dict and adds it to the payload so that the user doesn't have to do this manually. But let's sort out what we want it to look like first.

broker/setup_broker/lsst/setup_broker.sh
```diff
@@ -81,7 +83,10 @@ manage_resources() {
         bq --location="${region}" mk --dataset "${bq_dataset}"
 
         (cd templates && bq mk --table "${PROJECT_ID}:${bq_dataset}.${alerts_table}" "bq_${survey}_${alerts_table}_schema.json") || exit 5
+        (cd templates && bq mk --table "${PROJECT_ID}:${bq_dataset}.${supernnova_table}" "bq_${survey}_${supernnova_table}_schema.json") || exit 5
+        (cd templates && bq mk --table "${PROJECT_ID}:${bq_dataset}.${classifications_table}" "bq_${survey}_${classifications_table}_schema.json") || exit 5
```
troyraen: Let's revisit this table structure. I think it would be better for both us and users to have one wide table rather than separating things out into separate tables. 

I wonder if we can come up with a table structure and an alert structure that match so that we don't have to publish a separate topic for the bigquery subscriptions. I'm thinking of something that probably has a top-level field for each module, each with a set of fields nested underneath for its own data.

hernandezc1:
> Let me know what you find there.

@troyraen A summary of my progress so far:

- The base Python image (python:3.12-slim) lacked essential build tools like `gcc`, which caused the build to fail when trying to install `pandas` (`error: command 'gcc' failed: No such file or directory`)
- I updated the Dockerfile to install `gcc` and related build dependencies
- The Cloud Run service successfully deployed, but the service subsequently crashed with the following error message:
     - `ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject`
     - a relevant GitHub [discussion](https://github.com/Sinaptik-AI/pandas-ai/issues/1251) with the same issue.
     - it seems that the solution and next step is to pin compatible versions of `numpy` and `pandas`, redeploy, and see what happens

broker/cloud_run/lsst/classify_snn/main.py
hernandezc1: Haven't had the chance to confirm yet.

troyraen: I don't think it will. The reason we typically publish in json is because I didn't want to go through the hassle of creating a new avro schema for every module.

troyraen:
numpy 2.0 introduced many breaking changes, so that's my guess as to what's going on. I suggest figuring out which version of pandas we need (use the latest version unless supernnova or some other dependency requires something else) and then finding out whether that version of pandas needs numpy<2.0 or numpy>=2.0.

hernandezc1:
> numpy 2.0 introduced many breaking changes, so that's my guess as to what's going on. I suggest figuring out which version of pandas we need (use the latest version unless supernnova or some other dependency requires something else) and then finding out whether that version of pandas needs numpy<2.0 or numpy>=2.0.

SuperNNova [requires](https://github.com/supernnova/SuperNNova/blob/main/pyproject.toml) `pandas=1.5.1`. `numpy<2.0` is needed, and the current configuration (in my most recent commit) successfully deploys the service.

I was unable to successfully deploy the service using Python 3.12

broker/cloud_run/lsst/classify_snn/main.py
hernandezc1: You're right. Updated the serializer to `json`.

troyraen (approved):
Great!

PR #252, opened 2025-02-28
Add subshells to scripts

### LINKED ISSUE #240: Use a subshell when `cd`ing around in shell scripts
> The better way to do this is to use a subshell. Then you're not changing the implicit global state.
>
> ```
> cd templates
> bq mk --table ....
> ```
> 
> -> `(cd templates && bq mk --table ....) || exit 5`

_Originally posted by @tregle in https://github.com/mwvgroup/Pitt-Google-Broker/pull/232#discussion_r1761689055_

We have code in all of our shell scripts that `cd`s around like this. We should replace it as suggested above. This could probably be done by anyone who knows a little bash/shell even if they don't know anything about our repo or project.

hernandezc1:
Closes #240 (involves LSST, LVK, & ZTF resources). Updates ZTF deployment scripts to address changes in #253


troyraen:
Thanks for this, it will be very helpful! Can I ask you to make the directory reorg a separate PR? Since that one will touch a whole lot of files but will only rename them and not change the contents, making it separate will make things easier to review and to understand in the future when looking at the commit history. Here's how you can do that:

```sh
git switch develop
git pull
git switch -c u/ch/ztf/reorg  # or whatever you want to call the branch
git cherry-pick 007f27b
git push
```

Then open a new PR for that branch. Merge the new PR before merging this one. We ought to be able to do it quickly -- request my review and I'll make it a priority to look at. After that, I don't think that having commit `007f27b` still in this PR will cause any problems, but if it does then I'll show you how to remove it.

hernandezc1:
A few comments:

These changes were tested; there were no issues involved in deploying broker instances for LSST and LVK. However, there were some issues in deploying the Cloud Functions for ZTF. By default, the latest version of the google-cloud-sdk deploys Cloud Functions as Gen 2 Cloud Run Functions. To address this issue, I specified the `no-gen2` flag in each of the Cloud Functions' deployment scripts. 

Despite this, there seems to be an issue with the source code for `SuperNNova`, preventing it from being successfully deployed. I'm not sure whether that's within the scope of this PR and we should likely switch over to using Cloud Run anyways.

troyraen (approved):
I took a quick look and it looks good to me.

> there seems to be an issue with the source code for SuperNNova, preventing it from being successfully deployed. I'm not sure whether that's within the scope of this PR and we should likely switch over to using Cloud Run anyways.

Would the supernnova deployment that you did for elasticc be easy to port over? Maybe it's time to break things temporarily in service of making them better over the course of a few relatively quick PRs.

hernandezc1:
> Would the supernnova deployment that you did for elasticc be easy to port over? Maybe it's time to break things temporarily in service of making them better over the course of a few relatively quick PRs.

@troyraen Do you mean:
- merge supernnova deployment from `main/elasticc` -> `develop`? or
- update current supernnova module for ztf with supernnova deployment from elasticc?

troyraen:
> Do you mean: ...

I was agreeing with your comment that it would be good to switch the ZTF module to Cloud Run and meant the latter, "update current supernnova module for ztf with supernnova deployment from elasticc", thinking that might be the easiest approach. More generally, I'm voting in favor of giving the ZTF SuperNNova module a major upgrade rather than trying to fix the one that's here now. Looking at the current requirements.txt file I see that it's pinned to very old versions of both supernnova and pandas, so that could be causing problems as well.

troyraen:
Thanks! If ZTF supernnova module is still broken, open an issue for it to make sure we can track it until it's fixed.

PR #253, opened 2025-02-28
Update directory structure

hernandezc1:
Reorganizes repo. Creates `ztf` directory and stores appropriate files


hernandezc1:
It seems that in the process of moving files to the appropriate directory, the `krb5.conf` was automatically deleted. Not sure what to do about this @troyraen

troyraen (approved):
Thanks! 

For 'krb5.conf', just copy it from the develop branch and add it back in here. We'll lose the git history for that file but let's not worry about that. There are ways to get the file back that will preserve the history but in this case they're a bit involved and there's nothing particularly important to preserve anyway.

This will break things that rely explicitly on the old file paths. Looks like you'll be fixing most of them in #252. If there are others, we'll fix them as we find them. IMO, the benefits of this reorg outweigh the temporary breakages.

PR #254, opened 2025-03-04
Delete `check_cue_response` Cloud Function

### LINKED ISSUE #221: Delete the check_cue_response Cloud Function
> I think we can ignore this cloud function here. I'd like to just delete it. It used to do several checks on whether things were running properly and generate a notification/alarm if they weren't. But most (I think all) have become unnecessary over time. Some kind of check/alarm system might be useful, but I don't think there's anything about this one that we should salvage.
>
> _Originally posted by @troyraen in https://github.com/mwvgroup/Pitt-Google-Broker/pull/216#discussion_r1482598316_
            
I think the steps to delete it are:

- Delete its directory.
- Find and remove it from the docs.
- Delete the deployed instance. I don't think it's even been triggered in a long time, but we should check which topic it's connect to and delete that too if not being used.

hernandezc1:
Closes #221


hernandezc1:
@troyraen I haven't removed discussions of this Cloud Function from the docs in your 'working-notes'. Please let me know if you'd like me to.

troyraen (approved):
Thanks! A couple requests to clean up 'docs/source/broker/components/broker-schedule.rst'. Otherwise LGTM.

- The remaining content in the 'Alerting policy' section doesn't make much sense without the deleted content. If someone wants to set up a new one, I'm sure there are multiple ways and the one I used before is probably not the best one anyway. So let's just delete that whole section.
- The 'Auto-scheduler's Logs' is now empty, so delete the header as well.
- Update the table of contents at the top of the page.

PR #256, opened 2025-03-09
Use Cloud Build to create all resources

troyraen:
@hernandezc1 This is an example of how I envision using Cloud Build to create all resources needed for a given module (supernnova, in this case). Some notes:
- This is just a draft -- I have not tested it at all. It contains several `FIXME`s and `CHECKME`s, and it likely contains some bugs. 
- Example usage is at the top of 'deploy.sh'. Start looking there.
- It's up to you whether you want to actually merge this into your branch and then iterate on it, or if you want to leave it unmerged and just reference it for ideas when writing your own code. Makes no difference to me. I just wanted to show you what I had in mind. I'm assigning it to you so that it's clear that you can do what you want with it from here.

## Included changes

- Add 'create-ancillaries.sh' -- A dedicated script to create all ancillary resources (e.g., Pub/Sub topics and subscriptions). 
    - Add a step in 'cloudbuild.yaml' that executes this script.
    - Remove these calls from 'deploy.sh'.
- Add 'env.yaml' -- A file containing all configurable variables for the module. 
    - Update 'deploy.sh' to get variables from this yaml rather than from command line arguments.
- Add 'construct-name.sh' -- A script that takes a resource name stem as input and constructs the resource name by prepending/appending the survey name and testid as appropriate.
    - Update other scripts to use this whenever a resource name needs to be constructed.

## Known omissions (will need to be added sooner rather than later)

- [ ] No functionality to teardown/delete resources. Need to think about whether to create dedicated script(s) for this or add it to existing scripts.

## Future work

Turn 'deploy.sh', etc. into reusable scripts. Individual modules should define all resources that they rely on, including all configuration, in an 'env.yaml'. The reusable scripts should take that as input and be able to handle (eg., iterate over) whatever they find in there.

First steps will be:

- [ ] There are two `FIXME`s in 'deploy.sh' indicating that those blocks of code need to be taken out of that script. Goal is for that script to load env vars and launch cloud build, that's it.
- [ ] Define resource name stems in 'env.yaml' rather than in 'create-ancillaries.sh'.

The rest has moved to #264.

### DIFF
--- a/broker/cloud_run/lsst/classify_snn/cloudbuild.yaml
+++ b/broker/cloud_run/lsst/classify_snn/cloudbuild.yaml
@@ -1,29 +1,56 @@
-# https://cloud.google.com/build/docs/deploying-builds/deploy-cloud-run
-# containerize the module and deploy it to Cloud Run
-steps:
-# Build the image
-- name: 'gcr.io/cloud-builders/docker'
-  args: ['build', '-t', '${_REGION}-docker.pkg.dev/${PROJECT_ID}/${_REPOSITORY}/${_MODULE_IMAGE_NAME}', '.']
-# Push the image to Artifact Registry
-- name: 'gcr.io/cloud-builders/docker'
-  args: ['push', '${_REGION}-docker.pkg.dev/${PROJECT_ID}/${_REPOSITORY}/${_MODULE_IMAGE_NAME}']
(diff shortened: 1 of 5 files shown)


troyraen:
Opened #264 and updated the main comment.

PR #262, opened 2025-03-10
Update badges in README

troyraen:
- Add: Codacy
- Remove: python3.7, TravisCI


troyraen:
@djperrefort Ok to drop the TravisCI badge? Looks like TravisCI is not being used here at the moment; unless I'm missing something.

Issue #264, opened 2025-03-11
Create reusable deployment scripts


troyraen:
This is "Future work" that could follow #256. 

## Directory structure

Could look something like this (updated 13 Mar 2025):

```
broker/
├── deploy/
│   ├── configs-template.yaml       # env.yaml becomes this.
│   ├                               #   Modules will use it to define all resources and their
│   ├                               #   they rely on, including all required configurations.
│   ├── deploy.sh                   # Current classify_snn/deploy.sh generalizes to this.
│   ├                               #   Should be able to accept any cloud run module's configs.yaml as input.
│   ├── cloudbuild.cloud-run.yaml   # Every cloud run module will use this.
│   ├── cloudbuild.consumer.yaml    # For example.
│   ├── _create-ancillaries.sh       # Called by deploy.sh.
│   ├                               #   Should be able to iterate over entries found in
│   ├                               #   configs.yaml of a given type (eg, pubsub topic)
│   ├                               #   and create or delete them, agnostic to the details.
│   ├── _construct_name.sh          # Called by deploy.sh and others.
├── cloud_run/classify_snn/
│   └── main.py ... etc.            # All surveys use these classify_snn/ files.
│   ├── configs.lsst.yaml           # All surveys define their own configs file.
│   ├── configs.ztf.yaml
├── cloud_run/ps_to_storage/
│   └── main.py ... etc.            # All surveys use these ps_to_storage/ files.
│   ├── configs.lsst.yaml           # All surveys define their own configs file.
│   ├── configs.lvk.yaml
│   ├── configs.ztf.yaml
# ... etc...
```

## Steps

All of these are 'classify_snn/' --> 'deploy/':

- [ ] Move 'construct_name.sh' --> '_construct_name.sh'.
- [ ] Convert 'env.yaml' --> 'configs-template.yaml'.
- [ ] Convert 'deploy.sh' --> 'deploy.sh'.
- [ ] Convert 'cloudbuild.yaml' --> 'cloudbuild.cloud-run.yaml'.
- [ ] Generalize 'create-ancillaries.sh' --> '/_create-ancillaries.sh'.

## How this could be used with Terraform

[...]

PR #265, opened 2025-03-11
Use Cloud Run for ZTF's SuperNNova module

### LINKED ISSUE #255: Update ZTF SuperNNova module
The SuperNNova Cloud Function for ZTF currently fails to deploy as a result of outdated package dependencies. In addition to resolving these issues, we should transition to using Cloud Run instead of gen 1 Cloud Run Functions (previously known as Cloud Functions)

hernandezc1:
Fixes #255


hernandezc1:
Changes still need to be tested.

troyraen (commented):
Can lsst and ztf share one copy of the directory 'ZTF_DMAM_V19_NoC_SNIa_vs_CC_forFink/'? That would be a good step towards reusable modules. Maybe move it up two levels? That would put it in a weird place for a minute but it wouldn't be the end goal. #264. cloudbuild.yaml might (or might not) be a better place than the deploy.sh script to point at these files. If it helps, here is an example of how to copy scripts into the deployment -- not directly applicable but maybe something similar is:

https://github.com/mwvgroup/Pitt-Google-Broker/blob/ce60feb5d68c9fcd59962d9a5a86012739655bb2/broker/cloud_run/lsst/classify_snn/cloudbuild.yaml#L25

hernandezc1:
> Can lsst and ztf share one copy of the directory 'ZTF_DMAM_V19_NoC_SNIa_vs_CC_forFink/'?

@troyraen I don't think so (thinking long term here). It might be best to keep them separate for now since we will likely use different models (trained on different datasets) for each deployment of SuperNNova for a specific survey. The filenames produced as a result of training SuperNNova (e.g., `cli_args.json`, `data_form.json`, etc.) are not unique. It's know it's redundant how things are currently set up but this won't always be the case.

troyraen:
>  It's know it's redundant how things are currently set up but this won't always be the case.

Very good point.

troyraen (approved):
LGTM.

Too bad that we're not yet able to use BigQuery Subscriptions here. And sending data to two different tables compounds the problems we have at high alert rates. I'm thinking we should probably get rid of the "classifications" table. After a brief look, seems like the only thing in that table that is not captured in the others is the classifier version. We should add a version column to all of our value-added tables. The version should probably represent our module as a whole rather than just the version of, say, the SuperNNova python library we're using. But we'll need some way to keep track of what was changed in each version. Maybe we should start keeping CHANGELOGs for each module? 

If you have a few minutes this week, take a look at that ZTF classifications table and related modules (which I think includes tag and supernnova). Double check my assumptions from above and see if there's an easy way to make those modules send the combined info to the dedicated table only and stop publishing to that classifications table. Using a bigquery subscription would be great but, even without that, just making them publish to one table instead of two would be a big help. No need to spend a ton of time on it, and doesn't need to hold up this PR.

PR #267, opened 2025-04-08
Create `lite` module for LSST

hernandezc1:
This PR creates a `lite` module for LSST that publishes an alert stream containing a subset of the fields from the original alert packet. Currently, the fields contained in the lite alert are:

```
{'alertId': ,
 'diaSource': {'diaSourceId': ,
  'midpointMjdTai': ,
  'ra': ,
  'raErr': ,
  'dec': ,
  'decErr': ,
  'psfFlux': ,
  'psfFluxErr': ,
  'band': },
 'prvDiaSources': [{'diaSourceId': ,
   'midpointMjdTai': ,
   'ra': ,
   'raErr': ,
   'dec': ,
   'decErr': ,
   'psfFlux': ,
   'psfFluxErr': ,
   'band': }],
'prvDiaForcedSources': [{'diaForcedSourceId': ,
   'diaObjectId': ,
   'ra': ,
   'dec': ,
   'visit': ,
   'detector': ,
   'psfFlux': ,
   'psfFluxErr': ,
   'midpointMjdTai': ,
   'band': }],
'prvDiaNondetectionLimits': [{'ccdVisitId': ,
   'midpointMjdTai': ,
   'band': ,
   'diaNoise': }],
 'diaObject': {'diaObjectId': ,
  'nearbyObj1': ,
  'nearbyObj1Dist': ,
  'nearbyObj1LnP': ,
  'nearbyObj2': ,
  'nearbyObj2Dist': ,
  'nearbyObj2LnP': ,
  'nearbyObj3': ,
  'nearbyObj3Dist': ,
  'nearbyObj3LnP': ,
  'u_psfFluxErrMean': ,
  'g_psfFluxErrMean': ,
  'r_psfFluxErrMean': ,
  'i_psfFluxErrMean': ,
  'z_psfFluxErrMean': ,
  'y_psfFluxErrMean': },
'ssObject': {'ssObjectId': ,
  'firstObservationDate': ,
  'arc': ,
  'numObs': ,
  'MOID': ,
  'MOIDEclipticLongitude': ,
  'MOIDDeltaV': ,
  'u_H': ,
  'u_HErr': ,
  'u_Chi2': ,
  'g_H': ,
  'g_HErr': ,
  'g_Chi2': ,
  'r_H': ,
  'r_HErr': ,
  'r_HChi2': ,
  'i_H': ,
  'i_HErr': ,
  'i_HChi2': ,
  'z_H': ,
  'z_HErr': ,
  'z_HChi2': ,
  'y_H': ,
  'y_HErr': ,
  'y_HChi2': ,
  'medianExtendedness': ,}
}
```

### DIFF
--- a/broker/cloud_run/lsst/lite/main.py
+++ b/broker/cloud_run/lsst/lite/main.py
@@ -7,6 +7,7 @@ import os
 import flask
 import pittgoogle
 from google.cloud import logging
+from typing import Optional
 
 # [FIXME] Make this helpful or else delete it.
 # Connect the python logger to the google cloud logger.
@@ -105,7 +106,7 @@ def _create_lite_dict(alert_dict: dict, field_names: list[str]) -> dict:
 
 def _create_prv_sources_dict(
     source_history: list[dict], field_names: list[str]
-) -> list[dict] | None:
+) -> Optional[list[dict]]:
     """Create a list of prv_sources dictionaries."""
 
     if source_history is None:
--- a/broker/cloud_run/lsst/lite/main.py
+++ b/broker/cloud_run/lsst/lite/main.py
@@ -5,6 +5,7 @@
 
 import os
 from typing import Optional
+
 import flask
 import pittgoogle
 from google.cloud import logging
--- a/broker/cloud_run/lsst/lite/main.py
+++ b/broker/cloud_run/lsst/lite/main.py
@@ -4,10 +4,10 @@
 """This module creates a "lite" LSST alert containing a subset of fields from the original LSST alert."""
 
 import os
+from typing import Optional
 import flask
 import pittgoogle
 from google.cloud import logging
-from typing import Optional
 
 # [FIXME] Make this helpful or else delete it.
 # Connect the python logger to the google cloud logger.
@@ -107,7 +107,7 @@ def _create_lite_dict(alert_dict: dict, field_names: list[str]) -> dict:
 def _create_prv_sources_dict(
     source_history: list[dict], field_names: list[str]
 ) -> Optional[list[dict]]:
-    """Create a list of prv_sources dictionaries."""
+    """Create a list of prv_sources dictionaries if they exist."""
 
     if source_history is None:
         return source_history
--- a/broker/cloud_run/lsst/lite/main.py
+++ b/broker/cloud_run/lsst/lite/main.py
@@ -100,7 +100,7 @@ def _create_source_fields_list(alert: pittgoogle.Alert) -> list[str]:
         "dec_err",
         "flux",
         "flux_err",
-        "band",
+        "filter",
     ]
 
     return _get_survey_field_names(alert, broker_field_names)
--- a/broker/cloud_run/lsst/lite/main.py
+++ b/broker/cloud_run/lsst/lite/main.py
@@ -4,7 +4,6 @@
 """This module creates a "lite" alert containing a subset of fields from the original alert packet."""
 
 import os
-from typing import Optional
 
 import flask
 import pittgoogle
--- a/broker/cloud_run/lsst/lite/main.py
+++ b/broker/cloud_run/lsst/lite/main.py
@@ -118,7 +118,7 @@ def _create_fields_list(field: str) -> list[str]:
 
         return source_fields_list
 
-    elif field == "object":
+    if field == "object":
         object_fields_list = [
             "diaObjectId",
             "nearbyObj1",
@@ -140,7 +140,7 @@ def _create_fields_list(field: str) -> list[str]:
 
         return object_fields_list
 
-    elif field == "ssobject":
+    if field == "ssobject":
         ss_object_fields_list = [
             "ssObjectId",
             "firstObservationDate",
@@ -171,6 +171,9 @@ def _create_fields_list(field: str) -> list[str]:
         ]
 
         return ss_object_fields_list
+
+    else:
+        raise ValueError(f"Unrecognized field type: {field}")
 
 
 def _create_lite_dict(alert_dict: dict, field_names: list[str]) -> dict:
--- a/broker/cloud_run/lsst/lite/main.py
+++ b/broker/cloud_run/lsst/lite/main.py
@@ -67,14 +67,14 @@ def run():
 
 
 def _create_lite_alert(alert: pittgoogle.Alert) -> pittgoogle.Alert:
-    """Create a "lite" alert containing a subset of the fields of the original alert packet."""
+    """Creates a "lite" alert containing a subset of the fields of the original alert packet."""
 
-    # define the fields that will be present in the lite alert
-    source_fields_list = _create_fields_list(field="source")
-    object_fields_list = _create_fields_list(field="object")
-    ss_object_fields_list = _create_fields_list(field="ssobject")
+    # create a list of field names that will be included in a lite dictionary
+    source_fields_list = _get_fields(field_name=alert.get_key("source"))
+    object_fields_list = _get_fields(field_name=alert.get_key("object"))
+    ss_object_fields_list = _get_fields(field_name=alert.get_key("ss_object"))
 
-    # create lite versions of fields in the original alert packet
+    # create lite dictionaries for each of the fields that will be included in the lite alert
     object_lite_dict = _create_lite_dict(
         alert.dict.get(alert.get_key("object")), object_fields_list
     )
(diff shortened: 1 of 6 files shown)


broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,170 @@
...
+        "nearby_object3_distance",
+        "prob_is_object1",
+        "prob_is_object2",
+        "prob_is_object3",
+        "u_flux_err_mean",
+        "g_flux_err_mean",
+        "r_flux_err_mean",
+        "i_flux_err_mean",
+        "z_flux_err_mean",
+        "y_flux_err_mean",
```
troyraen: I'm not sure that we want to make generic names for all of these. The ugrizy fields in particular are quite specific to LSST -- some of those bands overlap with ZTF bands, but I'm not sure ZTF has equivalent fields that would map to these names. I could potentially be convinced to add the nearby-objects fields but I'm skeptical because, while those types of fields do tend to be included by different surveys, they don't quite have the same meanings across surveys and it'm not sure there will be a clear mapping between fields from different surveys and the ones you have listed here. For example, surveys might include a closest star and a closest galaxy and I'm not sure how to fit those in here. I'm even leery of a generic field like "nearby_galaxy1" because I'm concerned that it will obscure important differences in the meanings of those fields from different surveys. For example, nearby according to whom?... LSST might list the closest galaxy detected by LSST while ZTF might list the closest detected by PanSTARRS.

broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,170 @@
...
+
+def _get_survey_field_names(alert: pittgoogle.Alert, broker_field_names: list) -> list[str]:
+    """Returns a list of survey-specific field names to be included in the lite dictionary."""
+    _survey_field_names = [alert.get_key(field) for field in broker_field_names]
+
+    # fields may be lists, extract the second element
+    survey_field_names = [
+        _survey_field_name[1] if isinstance(_survey_field_name, list) else _survey_field_name
+        for _survey_field_name in _survey_field_names
+    ]
```
troyraen: What's the reason for extracting the second element? I'm getting a bit lost on what's happening between the different function calls and what the final result will look like.

broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,170 @@
...
+
+    return survey_field_names
+
+
+def _create_lite_dict(alert_dict: dict, field_names: list[str]) -> dict:
+    return _drop_fields(alert_dict, field_names)
```
troyraen: Can this function be removed and we just call `_drop_fields` directly?

broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,170 @@
...
+
+    prev_sources = []
+    for prv_s in source_history:
+        lite_source_dict = _drop_fields(prv_s, field_names)
+        prev_sources.append(lite_source_dict)
+
+    return prev_sources
+
+
+def _drop_fields(alert_dict: dict, field_names: list[str]) -> dict:
```
troyraen: I suggest dropping this function and just using line 168 in its place. If you do really want this to be a function, then I recommend changing the name because the signature makes me think that whatever is passed to `field_names` is going to be dropped.

troyraen (commented):
I left some thoughts below. Bigger picture, let's get more clear about what this schema should look like (conversation started in https://github.com/mwvgroup/pittgoogle-client/pull/91#discussion_r2044619852). Also, is this module also going to add attributes (several are added automatically by the client, but are those all that are needed for now?) and/or simple categorizations similar to the ZTF "tag" module?

broker/cloud_run/lsst/lite/main.py
hernandezc1: Here's an example:

```
broker_field_names = ["source", "ra", "dec"]
_survey_field_names = [alert.get_key(field) for field in broker_field_names]
print(_survey_field_names)

['diaSource', ['diaSource', 'ra'], ['diaSource', 'dec']]
```

As you can see, `get_key()` may sometimes return a list, so we want to extract the second element from the list.

```
survey_field_names = [
    _survey_field_name[1] if isinstance(_survey_field_name, list) else _survey_field_name
    for _survey_field_name in _survey_field_names
]
print(survey_field_names)

['diaSource', 'ra', 'dec']
```

broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,200 @@
...
+    ssobject_lite_dict = _create_lite_dict(
+        alert.dict.get(alert.get_key("ss_object")), ss_object_fields_list
+    )
+    # create the lite dictionary for the outgoing alert
+    alert_lite_dict = {
+        alert.get_key("alertid"): alert.get("alertid"),
+        alert.get_key("source"): source_lite_dict,
+        alert.get_key("prv_sources"): prev_sources_lite_dict,
+        "prvDiaForcedSources": alert.dict.get("prvDiaForcedSources"),
+        "prvDiaNondetectionLimits": alert.dict.get("prvDiaNondetectionLimits"),
```
troyraen:
```suggestion
        alert.get_key("prv_forced_sources"): alert.dict.get("prvDiaForcedSources"),
        alert.get_key("prv_nondetect_limits"): alert.dict.get("prvDiaNondetectionLimits"),
```

broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,200 @@
...
+def _create_lite_dict(alert_dict: dict, field_names: list[str]) -> dict:
+    """Returns a lite dictionary containing fields specified in field_names."""
+    if alert_dict is None:
+        return alert_dict
+    return {k: v for k, v in alert_dict.items() if k in field_names}
+
+
+def _create_prv_sources_lite_dict(
+    source_history: list[dict], field_names: list[str]
+) -> Optional[list[dict]]:
```
troyraen:
```suggestion
def _create_prv_sources_lite_dict(
    source_history: list[dict] | None, field_names: list[str]
) -> list[dict] | None:
```

broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,200 @@
...
+        alert.get_key("alertid"): alert.get("alertid"),
+        alert.get_key("source"): source_lite_dict,
+        alert.get_key("prv_sources"): prev_sources_lite_dict,
+        "prvDiaForcedSources": alert.dict.get("prvDiaForcedSources"),
+        "prvDiaNondetectionLimits": alert.dict.get("prvDiaNondetectionLimits"),
+        alert.get_key("object"): object_lite_dict,
+        alert.get_key("ss_object"): ssobject_lite_dict,
+    }
+
+    return pittgoogle.Alert.from_dict(payload=alert_lite_dict, schema_name=f"{SURVEY}")
```
troyraen: At minimum, pass along the attributes from the incoming alert. Are there other attributes you want to attach here?

troyraen (approved):
Thanks! A few minor comments below.

broker/cloud_run/lsst/lite/main.py
hernandezc1: @troyraen 
> Are there other attributes you want to attach here?

Not at the moment.

> At minimum, pass along the attributes from the incoming alert.

I think that might be repetitive with the current setup. As of right now, the attributes from the incoming alert are the same as `alert_lite` (i.e., if we choose to attach other attributes in the future then the incoming/outgoing alert attributes may differ) and these attributes are attached to outgoing Pub/Sub messages when the `publish` method is executed.

broker/cloud_run/lsst/lite/main.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _create_lite_dict(alert_dict: dict, field_names: list[str]) -> dict:
+    """Returns a lite dictionary containing fields specified in field_names."""
+    if alert_dict is None:
+        return alert_dict
+    return {k: v for k, v in alert_dict.items() if k in field_names}
+
+
+def _create_prv_sources_lite_dict(
```
hernandezc1, line 188: @troyraen Codacy had issues with the `|` operand:
```
unsupported operand type(s) for |
```
I've updated these lines to address the issue.

broker/cloud_run/lsst/lite/main.py
troyraen: There are attributes from the incoming Kafka stream that will be lost if not explicitly attached to this outgoing alert. Try it. I think you need to pass the incoming alert attributes to `Alert.from_dict()` to end up with the desired outcome (all incoming and added attributes preset in outgoing alert). If you do find that there are "added" attributes that don't end up in the outgoing alert, we may just want to de-indent [alert.py line 273](https://github.com/mwvgroup/pittgoogle-client/blob/9e2a4a00245a93d8530ec346773e0994667a648d/pittgoogle/alert.py#L273) so it's called every time `Alert.attributes` is.

broker/cloud_run/lsst/lite/main.py
troyraen: Using `typing.Optional` is fine. It shouldn't be necessary as of python 3.9(?) but sometimes it can be tricky to get the syntax right with `|`.

broker/cloud_run/lsst/lite/main.py
hernandezc1: > There are attributes from the incoming Kafka stream that will be lost if not explicitly attached to this outgoing alert. 

You're absolutely right. Thank you for the reminder. I'll pass them to outgoing messages

PR #272, opened 2025-06-06
Update `lsst-lite` module

hernandezc1:
This PR addresses the following warning that gets printed when an outgoing message from this module is published:

```
No alert_bytes provided. Cannot infer schema version. Methods that rely on it will be unavailable.
```

**Summary of changes**:
- Outgoing Alert objects have the argument `schema_name` set to: `default`
- Outgoing messages now contain a top-level key: `alert_lite`; the value is the lite alert


broker/cloud_run/lsst/lite/main.py
```diff
@@ -60,7 +60,7 @@ def run():
     except pittgoogle.exceptions.BadRequest as exc:
         return str(exc), HTTP_400
 
-    TOPIC_LITE.publish(_create_lite_alert(alert), serializer="json")
+    TOPIC_LITE.publish(_create_lite_alert(alert))
```
troyraen, line 63: Don't we want the outgoing alert to be json?

broker/cloud_run/lsst/lite/main.py
```diff
@@ -98,7 +98,9 @@ def _create_lite_alert(alert: pittgoogle.Alert) -> pittgoogle.Alert:
     }
 
     return pittgoogle.Alert.from_dict(
-        payload=alert_lite_dict, schema_name=f"{SURVEY}", attributes={**alert.attributes}
+        payload={"alert": alert_lite_dict},
+        schema_name="default",
```
troyraen:
```suggestion
```

Does it also work smoothly if this kwarg is left out altogether? I think it should, but maybe something I'm not thinking of. Curious largely because I wonder if the options `"default"` and `None` (which is the default value for both the from_dict() method and the Alert class itself) are actually completely redundant and we should get rid of one of them to avoid confusion.

troyraen (approved):
Two questions below, but no blockers.

broker/cloud_run/lsst/lite/main.py
hernandezc1: > Does it also work smoothly if this kwarg is left out altogether?

@troyraen I just checked and yes, it still works smoothly if we leave the kwarg out.

broker/cloud_run/lsst/lite/main.py
hernandezc1: @troyraen If the argument `schema_name` is set to the value `default`, the serializer will be JSON by default as well

troyraen: Oh right, thanks.

broker/cloud_run/lsst/lite/main.py
troyraen: Great, I made a note to myself in mwvgroup/pittgoogle-client#88 to try removing `"default"`.

PR #273, opened 2025-06-09
Add variable star classifier (UPSILoN) to LSST Science Pipeline

hernandezc1:
This PR introduces a new module to Pitt-Google's LSST science pipeline that uses [UPSILoN](https://github.com/dwkim78/upsilon) to identify variable sources and classify them.

**Summary of changes**:
- Outgoing messages from the `lsst-lite` module triggers a Cloud Run service (`lsst-upsilon`) to classify periodic variable stars
- Outgoing alerts are published to a Pub/Sub topic: `lsst-upsilon`
     - Messages published to this topic are organized using two top level keys: `alert_lite` and `upsilon`. 
     - The `alert_lite` value contains the original fields of the lite alert packet; the `upsilon` value contains three outputs generated by UPSILoN for each band used to observe the diaObject: _label_, _probability_, and _flag_ as well as the diaObjectId and diaSourceId associated with the alert packet.
     - New Pub/Sub message attributes are included when these messages are published to the `lsst-upsilon` topic: 
          - `pg_upsilon_u_label`, `pg_upsilon_u_flag`
          - `pg_upsilon_g_label`, `pg_upsilon_g_flag`
          - `pg_upsilon_r_label`, `pg_upsilon_r_flag`
          - `pg_upsilon_i_label`, `pg_upsilon_i_flag`
          - `pg_upsilon_z_label`, `pg_upsilon_z_flag`
          - `pg_upsilon_y_label`, `pg_upsilon_y_flag`
- A BigQuery subscription attached to the `lsst-upsilon` topic writes the `upsilon` data to a BigQuery table: `upsilon` (dataset: `lsst_value_added`).


hernandezc1:
> New Pub/Sub message attributes are included when these messages are published to the lsst-upsilon topic

Here is a [link that defines the outputs generated by UPSILoN](https://github.com/dwkim78/upsilon?tab=readme-ov-file#classification).

`label` is a `STRING`; `probability` is a `FLOAT`; and `flag` is an `INTEGER`.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env python3
+# -*- coding: UTF-8 -*-
+
+"""Classify alerts using UPSILoN (Kim & Bailer-Jones 2015)."""
+
+import os
+from typing import Dict
```
troyraen:
```suggestion
```

Can just use the builtin `dict` with recent-ish versions of python.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+                "pg_upsilon_y_flag": upsilon_dict["y_flag"],
+            },
+            schema_name="default",
+        )
+    )
+
+    return "", HTTP_204
+
+
+def _classify_with_UPSILoN(alert_lite: pittgoogle.Alert) -> Dict:
```
troyraen:
```suggestion
def _classify_with_UPSILoN(alert_lite: pittgoogle.Alert) -> dict:
```

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+        # classify
+        label, probability, flag = upsilon.predict(rf_model, features)
+        outgoing_dict[f"{band}_label"] = label
+        outgoing_dict[f"{band}_probability"] = probability
+        outgoing_dict[f"{band}_flag"] = flag
+
+    return outgoing_dict
+
+
+def _create_dataframe(alert_dict: pittgoogle.Alert) -> "pd.DataFrame":
```
troyraen:
```suggestion
def _create_dataframe(alert_dict: pittgoogle.Alert) -> pd.DataFrame:
```

No need to quote since you're doing a normal import of pandas.

broker/cloud_run/lsst/classify_upsilon/requirements.txt
```diff
@@ -0,0 +1,17 @@
+# As explained here
+# https://cloud.google.com/functions/docs/writing/specifying-dependencies-python
+# dependencies for a Cloud Function must be specified in a `requirements.txt`
+# file (or packaged with the function) in the same directory as `main.py`
+
+google-cloud-logging
+pittgoogle-client>=0.3.15
+upsilon
+scikit-learn==1.2.2
+numpy==1.26.4
```
troyraen: When either pinning to a specific version or upper-limiting a version, comment as to why. Those types of dependency requirements often need to be revisited in the future and the comment can save hours of digging. (Less important for lower limits, eg pittgoogle-client, because they don't tend to become headaches in the future.)

If you can relax these version requirements, that tends to make the code easier to maintain. It's probably worth it to spend a little bit of time checking, but definitely not worth a lot of time. If you don't know what to check, here's my train of thought...

numpy v2.0 included major breaking changes, and this had widespread knock-on effects since so many other libraries depend on numpy. So my guess is that upsilon still depends on numpy<2.0 (and/or maybe pittgoogle-client does? I hope not, but don't remember atm) but scikit-learn upgraded to numpy>2.0 recently. So I recommend trying:

```
scikit-learn
numpy<2.0
```

and let `pip` decide which scikit-learn version to use.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+HTTP_400 = 400  # HTTP code: Bad Request
+
+# GCP resources used in this module
+TOPIC = pittgoogle.Topic.from_cloud("upsilon", survey=SURVEY, testid=TESTID, projectid=PROJECT_ID)
+
+app = flask.Flask(__name__)
+
+
+@app.route(ROUTE_RUN, methods=["POST"])
+def run():
```
troyraen:
```suggestion
def run() -> tuple[str, int]:
```

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+                "pg_upsilon_i_label": upsilon_dict["i_label"],
+                "pg_upsilon_i_flag": upsilon_dict["i_flag"],
+                "pg_upsilon_z_label": upsilon_dict["z_label"],
+                "pg_upsilon_z_flag": upsilon_dict["z_flag"],
+                "pg_upsilon_y_label": upsilon_dict["y_label"],
+                "pg_upsilon_y_flag": upsilon_dict["y_flag"],
+            },
+            schema_name="default",
+        )
+    )
```
troyraen: Update our [Data Listings](https://mwvgroup.github.io/pittgoogle-client/listings.html) page in the client docs so that users know about this stream and what fields to expect (both payload and attributes) and how they should interpret, eg, "pg_upsilon_y_flag". It would be great to include a full list and descriptions of the message attributes  (not currently there for any stream) so that folks know what's available for filtering. We may want to reorganize that page to make room. Maybe we should move the actual data listings (the tables) to the suvery-specific pages (like [LSST](https://mwvgroup.github.io/pittgoogle-client/surveys/lsst.html#survey-lsst))? Anyway, just thinking out loud and that bit is not urgent, but it would be great to at least do the simple thing and list this stream in the current format.

There's also the actual broker docs... we really should have docs for this repo (we do) that are up to date (we don't). I encourage you to consider that, especially for the sake of the developers who will come after you. Just some quick and basic documentation describing the pipelines/modules you're adding would be good. I probably put too much and/or not the right kind of detail in the pages I wrote, so don't feel bound to imitate that.  Also, what I currently have there is probably not well organized. If you want to propose a re-org, I'm definitely open to that. But again, priority = data listings; simple updates to broker docs as you go would be helpful.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+    outgoing_dict = {
+        "diaObjectId": alert_lite_dict["diaObject"]["diaObjectId"],
+        "diaSourceId": alert_lite_dict["diaSource"]["diaSourceId"],
+    }
+
+    # load UPSILoN's classification model
+    rf_model = upsilon.load_rf_model()
+    for band in bands:
+        # define parameters for feature extractions
+        filter_diaSources = alert_df[alert_df["band"] == band]
```
troyraen, line 104:
```suggestion
    for band, filter_diaSources in alert_df.groupby("band"):
```

What you're doing is fine. Suggesting this alternate because I find it more readable and intuitive, making it immediately clear which slice of data is being used during a given iteration.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+        mask = filter_diaSources["psfFlux"].to_numpy() > 0
```
troyraen: Maybe rename `mask` to something like `flux_filter` or `flux_gt_zero`? "Mask" typically means that `True` elements should be discarded, opposite of what's meant here.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+    https://github.com/lsst/tutorial-notebooks/blob/044219c9ae5521edcc816af88e4b341e19326dbf/DP0.2/01_Introduction_to_DP02.ipynb#L511
+
+    Converts flux [nJy] to AB magnitude.
+    """
+    return -2.50 * np.log10(flux) + 31.4
+
+
+def _calculate_mag_err(flux: np.ndarray, flux_err: np.ndarray) -> float:
+    """Calculates magnitude uncertainty."""
+    return (-1.08574 / flux) * flux_err
```
troyraen: Where does this formula come from? Also, does this really return a float or is it an array?

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+    alert_lite_dict = alert_lite.dict["alert_lite"]
+    alert_df = _create_dataframe(alert_lite_dict)
+    bands = ["u", "g", "r", "i", "z", "y"]
+    outgoing_dict = {
+        "diaObjectId": alert_lite_dict["diaObject"]["diaObjectId"],
+        "diaSourceId": alert_lite_dict["diaSource"]["diaSourceId"],
+    }
+
+    # load UPSILoN's classification model
+    rf_model = upsilon.load_rf_model()
```
troyraen: Maybe this should be done once at the top of the module? Assuming it's the same for every alert.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+    try:
+        alert_lite = pittgoogle.Alert.from_cloud_run(envelope, "default")
+    except pittgoogle.exceptions.BadRequest as exc:
+        return str(exc), HTTP_400
+
+    upsilon_dict = _classify_with_UPSILoN(alert_lite)
+
+    TOPIC.publish(
+        pittgoogle.Alert.from_dict(
+            {**alert_lite.dict, "upsilon": upsilon_dict},
```
troyraen:
```suggestion
            {"alert_lite": alert_lite.dict, "upsilon": upsilon_dict},
```

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    # extract the envelope from the request that triggered the endpoint
+    # this contains a single Pub/Sub message with the alert to be processed
+    envelope = flask.request.get_json()
+    try:
+        alert_lite = pittgoogle.Alert.from_cloud_run(envelope, "default")
+    except pittgoogle.exceptions.BadRequest as exc:
+        return str(exc), HTTP_400
+
+    upsilon_dict = _classify_with_UPSILoN(alert_lite)
```
troyraen: There's nothing wrong with this function name per se, but in python it's pretty standard for function names to be snake case (so, all lower case). Classes use upper camel case, functions and variables use snake case, and constants use all caps. These are not official rules but widely adopted conventions. The advantage of following them is that the code becomes more understandable.

troyraen (approved):
Looks good! Top three things are:

- I think a top-level "alert_lite" key needs to be added to the outgoing message.
- Update our Data Listings page (client docs) with the new stream and table info.
- Spend a little time with the requirements. Try to relax pinned versions to upper limits (or no restriction). Comment why versions must be restricted.

My other suggestions are mostly stylistic FYIs.

broker/cloud_run/lsst/classify_upsilon/main.py
hernandezc1: Thank you for catching this!

broker/cloud_run/lsst/classify_upsilon/main.py
hernandezc1: I've updated the function. It returns an array.

> Where does this formula come from?

I derived it:
$m = -2.5 \log_{10}(f) + 31.4$

$\sigma_{m} = \left| \frac{\partial m}{\partial f} \right| \sigma_{f} = \left| \frac{\partial}{\partial f} \left(-2.5 \log_{10}(f) + 31.4\right) \right| \sigma_{f}$

$\sigma_{m} = \left| \frac{-2.5}{\ln(10) f} \right| \sigma_{f} = \left| \frac{-2.5(0.43429...)}{f} \right| \sigma_{f}$

$\therefore \sigma_m = \frac{1.08574}{f} \sigma_f$

hernandezc1: The function now returns:

```
abs(-2.5 / (flux * np.log(10))) * flux_err
```

Issue #172, opened 2022-09-26
Replace the Metadata Collector with BigQuery subscriptions


troyraen:
- Motivated by #171 and work currently in progress.
- Proposes a more fundamental solution to the problems that #92 attempts to address, and has a broader impact.
- Proposes a workflow that would allow us to simply **_remove_** all of the BigQuery API calls that #96 (specifically [this comment](https://github.com/mwvgroup/Pitt-Google-Broker/issues/96#issuecomment-929401416)) identifies as majorly problematic. 

### Overview

This issue proposes an entirely new workflow for the broker's metadata collection process. It also documents our reasons for collecting specific metadata to begin with. #171 details fundamental problems with the current system design.

The new workflow can potentially solve many problems at once, at least two of which will be significant barriers to processing at LSST scale if left unaddressed. It is based on a new Pub/Sub feature:  [BigQuery subscriptions](https://cloud.google.com/pubsub/docs/bigquery?_ga=2.31545015.-865460638.1622387917#properties_subscription).

Here is an [architecture diagram](https://pitt-broker.readthedocs.io/en/latest/_images/architecture-current.svg) for reference (especially in the section outlining the new workflow).

### Motivation for a metadata collector

Enable the following project-level goals:

- Track data provenance. Specifically, the broker and classifier versions that processed each alert.
- Track pipeline performance. Specifically: module-level processing times, delay times, dropped alerts, duplicated alerts, etc.

### Outline of new workflow

[BigQuery subscriptions](https://cloud.google.com/pubsub/docs/bigquery?_ga=2.31545015.-865460638.1622387917#properties_subscription) is a new feature in Pub/Sub that basically let you point a **_push subscription_** directly at a **_BigQuery table_**. It includes the following important options, which make the new workflow possible: 1) write both the message _data_ and _metadata_ to the table, and 2) automatically drop fields that are present in the message but _not_ the table schema. 

Basic envisioned workflow: 

1. Each module will publish one output stream (i.e., topic) that contains:
    1) the results of its processing, 
    2) relevant metadata, and 
    3) (usually) a copy of the incoming data/alert. 
2. Each module's topic will have a BigQuery subscription attached. The subscription will automatically send the message to a BigQuery table, dropping any data/metadata that is not in the table's schema. 
    - Things like the original alert data can be included in every message, but only stored in a single table. 
    - Similar modules (like classifiers) can all push results to the same table.
[...]

PR #277, opened 2025-06-17
Ingest Swift/BAT-GUANO Alert Stream

hernandezc1:
This PR creates the necessary resources to ingest the [JSON-Serialized GCN Notice Types](https://gcn.nasa.gov/missions/swift#json-serialized-gcn-notice-types-in-gcn-kafka) generated by the Neil Gehrels Swift Observatory. Alerts are published to a Pub/Sub topic: `swift-alerts`. A BigQuery subscription attached to the topic writes the alert data to a BigQuery table (`alerts_v4_5_0`) in the dataset `swift_alerts`


troyraen:
Yay!

One thought: We already listen to one GCN stream (LVK) and may want to add more in the future. Consider how we can make this easy for ourselves. For example, maybe all of our GCN consumers can use the same base container image and ingest different streams based only on an environment variable (and maybe separate credentials)? I don't know the answer or even all the possibilities, just putting a bug in your ear.

broker/consumer/swift/ps-connector.properties
```diff
@@ -0,0 +1,41 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# --------------------------------------------------------------------------
+
+# Unique name for the Pub/Sub sink connector.
+name=CPSSinkConnector
+# Tha Java class for the Pub/Sub sink connector.
+connector.class=com.google.pubsub.kafka.sink.CloudPubSubSinkConnector
+# The maximum number of tasks that should be created for this connector.
+tasks.max=1
```
troyraen, line 30: What is the alert rate you expect from this stream? I'm guessing that one task is probably fine, but it's worth checking.

broker/consumer/swift/vm_install.sh
```diff
@@ -0,0 +1,77 @@
+#! /bin/bash
+# Installs the software required to run the Kafka Consumer.
+# Assumes a Debian 10 OS.
```
troyraen: Double check this. I think you're using Debian 11.

broker/consumer/swift/vm_install.sh
```diff
@@ -0,0 +1,77 @@
...
+#--- Install general utils
+apt-get update
+apt-get install -y wget screen software-properties-common snapd
+# software-properties-common installs add-apt-repository
+# install yq (requires snap)
+snap install core
+snap install yq
+
+#--- Install Java and the dev kit
+# see https://www.digitalocean.com/community/tutorials/how-to-install-java-with-apt-on-debian-10
```
troyraen: Similar to above. I think this should be a Debian 11 link.

broker/consumer/swift/vm_install.sh
```diff
@@ -0,0 +1,77 @@
...
+# install
+apt-get update && sudo apt-get install -y confluent-platform
+echo "Done installing Confluent Platform."
+
+#--- Install Kafka -> Pub/Sub connector
+# see https://github.com/googleapis/java-pubsub-group-kafka-connector/tree/main
+echo "Installing the Kafka -> Pub/Sub connector"
+(
+    plugindir=/usr/local/share/kafka/plugins
+    CONNECTOR_RELEASE="1.1.0"
```
troyraen: This version is more than 2 yrs old now. Latest version is 1.3.2, released 2025-06-02. It's probably worth a quick test to see if we can upgrade to the latest version without much trouble. If it turns into a rabbit hole, ok to punt that task and continue with this version for now.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -0,0 +1,140 @@
...
+            testid_suffix="-${testid}"
+        fi
+    fi
+
+    echo "${base_name}${testid_suffix}"
+}
+
+#--- GCP resources used directly in this script
+broker_bucket=$(define_GCP_resources "${PROJECT_ID}-${survey}-broker_files")
+bq_dataset_alerts=$(define_GCP_resources "${survey}_alerts")
```
troyraen:
```suggestion
bq_dataset_alerts=$(define_GCP_resources "${survey}")
```

Started this conversation on slack... Our datasets for other surveys are just named as the name of the survey. I see that we've started putting our value-added products in a separate dataset but I think that only creates more confusion than it's worth.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -0,0 +1,140 @@
...
+    fi
+
+    if [ "$mode" = "setup" ]; then
+        # create BigQuery dataset and table
+        echo
+        echo "Creating BigQuery dataset and table..."
+        bq --location="${region}" mk --dataset "${bq_dataset_alerts}"
+
+        (cd templates && bq mk --table "${PROJECT_ID}:${bq_dataset_alerts}.${alerts_table}" "bq_${survey}_${alerts_table}_schema.json") || exit 5
+        bq update --description "Alert data from Swift/BAT-GUANO. This table is an archive of the swift-alerts Pub/Sub stream. It has the same schema as the original alert bytes, including nested and repeated fields." "${PROJECT_ID}:${bq_dataset_alerts}.${alerts_table}"
```
troyraen: Add a link to the survey's documentation of the schema or some other page where users can find more info, if you know it.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -0,0 +1,140 @@
...
+define_GCP_resources() {
+    local base_name="$1"
+    local testid_suffix=""
+
+    if [ "$testid" != "False" ]; then
+        if [ "$base_name" = "${survey}_alerts" ]; then
+            testid_suffix="_${testid}"  # complies with BigQuery naming conventions
+        else
+            testid_suffix="-${testid}"
+        fi
```
troyraen: I recommending finding a more clear and direct way to identify whether "-" or "_" should be used here. Maybe a boolean function arg? Using the resource name makes this code quite brittle because we have to remember to change the guts of the function if we change the resource name and/or add new resources.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -0,0 +1,140 @@
...
+        # create Pub/Sub
+        echo
+        echo "Configuring Pub/Sub resources..."
+        gcloud pubsub topics create "${topic_alerts}"
+        gcloud pubsub topics create "${deadletter_topic_bigquery_import}"
+        gcloud pubsub subscriptions create "${deadletter_subscription_bigquery_import}" --topic="${deadletter_topic_bigquery_import}"
+        gcloud pubsub subscriptions create "${subscription_alerts_reservoir}" --topic="${topic_alerts}"
+        # in order to create BigQuery subscriptions, ensure that the following service account:
+        # service-<project number>@gcp-sa-pubsub.iam.gserviceaccount.com" has the
+        # bigquery.dataEditor role for each table
```
troyraen: Can the code for this be included here? Ideally, the setup script(s) should take care of all setup and the developer should not have to do things like this manually. If there is a good reason for the developer to do something manually, those instructions should be included with the other setup instructions rather than being buried in code comments.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -0,0 +1,140 @@
...
+            --use-table-schema \
+            --drop-unknown-fields \
+            --dead-letter-topic="${deadletter_topic_bigquery_import}" \
+            --max-delivery-attempts=5 \
+            --dead-letter-topic-project="${PROJECT_ID}"
+
+        # set IAM policies on resources
+        user="allUsers"
+        roleid="projects/${GOOGLE_CLOUD_PROJECT}/roles/userPublic"
+        gcloud pubsub topics add-iam-policy-binding "${topic_alerts}" --member="${user}" --role="${roleid}"
```
troyraen: Make the bigquery dataset public as well.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -0,0 +1,140 @@
...
+            o="GSUtil:parallel_process_count=1" # disable multiprocessing for Macs
+            gsutil -m -o "${o}" rm -r "gs://${broker_bucket}"
+            bq rm -r -f "${PROJECT_ID}:${bq_dataset_alerts}"
+            gcloud pubsub topics delete "${topic_alerts}"
+            gcloud pubsub topics delete "${deadletter_topic_bigquery_import}"
+            gcloud pubsub subscriptions delete "${subscription_alerts_reservoir}"
+            gcloud pubsub subscriptions delete "${deadletter_subscription_bigquery_import}"
+            gcloud pubsub subscriptions delete "${subscription_bigquery_import}"
+        fi
+    fi
```
troyraen, line 174: 1) Add a comment explaining that the consumer VM will be deleted by the create_vm.sh script. 

2) Add a print statement for the case `environment_type="production"` explaining why nothing happens and what the developer needs to do instead. You can copy/paste the one from the ZTF setup_broker.sh.

troyraen (changes requested):
Thanks, this will be great! A few things to clarify below.

I assume this is consumer will be one that we'll leave running 24/7?

I strongly encourage you to include the storage of alerts in a bucket with this PR. A bucket is more valuable to us than a bigquery table. We can create or recreate a table from alerts in a bucket but it's difficult to create a bucket of alerts from a bigquery table.  It would also be easy to "replay" a pubsub stream using alerts in a bucket but that would be difficult if trying to use a table. We've already run into this problem with LSST (and/or LVK?) when we wanted to change the table but had trouble because we didn't have a bucket to draw from.

broker/consumer/swift/ps-connector.properties
troyraen: Now I see in the data listings PR, "less than 1 alert per week". So 1 task will be fine.

broker/setup_broker/swift/setup_broker.sh
troyraen: Actually, maybe we should link to one of our pages instead. Maybe the survey page at docs/source/surveys/swift.rst being added in https://github.com/mwvgroup/pittgoogle-client/pull/100?

broker/consumer/swift/vm_install.sh
hernandezc1: @troyraen it's actually using Debian 12! Will update the comments in this script. I was reusing code from previous vm_install scripts (e.g., LVK).

broker/setup_broker/swift/templates/bq_swift_alerts_v4_5_0_schema.json
```diff
@@ -0,0 +1,266 @@
...
+        "description": "Position angle of semi-major axis in instrument coordinates [deg]."
+    },
+    {
+        "name": "healpix_url",
+        "type": "STRING",
+        "mode": "NULLABLE",
+        "description": "URL of HEALPix localization probability file."
+    },
+    {
+        "name": "healpix_file",
```
hernandezc1, line 141: @troyraen do we want to keep the healpix_file field here? This information will be available in the GCS bucket. We had a similar conversation when we first started storing alert data in BigQuery for LVK

broker/setup_broker/swift/setup_broker.sh
hernandezc1: Yes, I've now included the code here

broker/setup_broker/swift/setup_broker.sh
hernandezc1: The process of making the dataset public involves a few more steps than I was expecting. The documentation I referenced can be accessed [here](https://cloud.google.com/bigquery/docs/control-access-to-resources-iam#grant_access_to_a_dataset)

broker/setup_broker/swift/setup_broker.sh
hernandezc1: The function has been updated

broker/consumer/swift/vm_install.sh
troyraen: > it's actually using Debian 12! 

even better!

broker/setup_broker/swift/setup_broker.sh
troyraen: I recall it being more difficult for bigquery than pubsub as well. Looks like you worked it out, so great job!

broker/setup_broker/swift/templates/bq_swift_alerts_v4_5_0_schema.json
troyraen: My default instinct is to not alter the survey's schema. And/but if we are serving a copy of this file from GCP, we should definitely figure out how/where to advertise that since getting it from within GCP will be more convenient for at least some users. 

I'm not sure I  remember the details about what's going on here... Is this the file containing the probability that the event originated from a particular sky location (ie, a probability per healpix pixel)? And are we ingesting that file and serving it from a GCS bucket in addition to the alerts? If so, maybe we leave this original schema field intact but add another field called something like "healpix_file_gcp" or "healpix_file_pg" or "healpix_file_cloud"? I'm much more comfortable adding a new field than changing the existing one. Happy to talk this through in more detail on Thursday if I'm missing the point here.

broker/cloud_run/swift/ps_to_storage/main.py
```diff
@@ -0,0 +1,107 @@
...
+ROUTE_RUN = "/"  # HTTP route that will trigger run(). Must match deploy.sh
+
+# Variables for outgoing data
+HTTP_204 = 204  # HTTP code: Success
+HTTP_400 = 400  # HTTP code: Bad Request
+
+# GCP resources used in this module
+TOPIC_ALERTS_JSON = pittgoogle.Topic.from_cloud(
+    "alerts-json", survey=SURVEY, testid=TESTID, projectid=PROJECT_ID
+)
```
troyraen, line 36: Do these alerts come from Swift as json or avro? If json, I think the name of this topic can just be `swift-alerts` since we're not changing the serialization. Also, if this stays as `swift-alerts-json` does that mean we don't publish any topic that's just called `swift-alerts`? I think from a usability/consistency standpoint we should always publish a stream called `<survey>-alerts` that is just a pass through of the survey's full (but deduplicated) alert stream from Kafka (or whatever) into Pub/Sub.

If that seems confusing in comparison with our LSST streams where `lsst-alerts` is avro and `lsst-alerts-json` is json, maybe we consider changing those names so that `lsst-alerts` is the json version and we make the avro one called `lsst-alerts-avro`? Benefit of the current naming is that `lsst-alerts` is byte-for-byte the same as what Rubin publishes. That was my original intention for all of our `<survey>-alerts` streams, and in that sense using `swift-alerts` is consistent (assuming Swift really does publish these as json -- otherwise, sorry for this irrelevant tangent). But since all of our topics downstream of `<survey>-alerts` use json exclusively, I can see an argument for making all `<survey>-alerts` streams json as well.

broker/cloud_run/swift/ps_to_storage/deploy.sh
troyraen, line 1: There are a bunch of variables in here with names that include "avro" -- I'm sure this is a holdover from the ZTF and LSST modules. For reusability, we should change those names because not all surveys publish avro alerts. For Swift in particular, assuming they publish in json and not avro, having "avro" here is confusing.

broker/cloud_run/swift/ps_to_storage/deploy.sh
```diff
@@ -0,0 +1,95 @@
...
+    fi
+    echo "${base_name}${testid_suffix}"
+}
+
+#--- GCP resources used in this script
+artifact_registry_repo=$(define_GCP_resources "${survey}-cloud-run-services")
+cr_module_name=$(define_GCP_resources "${survey}-${MODULE_NAME}")  # lower case required by cloud run
+gcs_avro_bucket=$(define_GCP_resources "${PROJECT_ID}-${survey}_alerts")
+ps_input_subscrip=$(define_GCP_resources "${survey}-alerts_raw") # pub/sub subscription used to trigger cloud run module
+ps_subscription_avro=$(define_GCP_resources "${survey}-alert_avros-counter")
```
troyraen: Are you actually using these counter subscriptions? If not, get rid of this. I think ZTF is the only broker pipeline that uses them. 

FWIW, ZTF uses them with an ancillary module that helps us track broker performance. The module very useful, but the way I wrote it is complicated and makes it expensive and that's why we haven't added it to any other broker pipeline. We should rewrite it using bigquery subscriptions (which didn't exist when I wrote the original) -- https://github.com/mwvgroup/Pitt-Google-Broker/issues/172.

broker/cloud_run/swift/ps_to_storage/deploy.sh
```diff
@@ -0,0 +1,95 @@
...
+ps_topic_avro=$(define_GCP_resources "projects/${PROJECT_ID}/topics/${survey}-alert_avros")
```
troyraen: I know this is the original name of the topic, but using "avros" is not generically good for the aforementioned reasons, and if Swift publishes json then we definitely shouldn't use it here. Let's think of something better. `<survey>-alert_in_bucket`? Just off the top of my head.

broker/cloud_run/swift/ps_to_storage/deploy.sh
```diff
@@ -0,0 +1,95 @@
...
+    echo
+    echo "Creating trigger subscription for ${MODULE_NAME} Cloud Run service..."
+    # WARNING:  This is set to retry failed deliveries. If there is a bug in main.py this will
+    # retry indefinitely, until the message is delete manually.
+    gcloud pubsub subscriptions create "${ps_input_subscrip}" \
+        --topic "${ps_trigger_topic}" \
+        --topic-project "${PROJECT_ID}" \
+        --ack-deadline=600 \
+        --push-endpoint="${url}${ROUTE_RUN}" \
+        --push-auth-service-account="${runinvoker_svcact}"
```
troyraen: Did you figure out how to limit the number of retries for cloud run services? If so, implement it here so we can drop this disaster waiting to happen 🥴.

broker/cloud_run/swift/ps_to_storage/main.py
```diff
@@ -0,0 +1,107 @@
...
+    bucket_name = f"{bucket_name}-{TESTID}"
+
+client = storage.Client()
+bucket = client.get_bucket(client.bucket(bucket_name, user_project=PROJECT_ID))
+
+app = flask.Flask(__name__)
+
+
+@app.route(ROUTE_RUN, methods=["POST"])
+def run():
```
troyraen:
```suggestion
def run() -> tuple[str, int]:
```

broker/cloud_run/swift/ps_to_storage/main.py
```diff
@@ -0,0 +1,107 @@
...
+    try:
+        alert = pittgoogle.Alert.from_cloud_run(envelope, schema_name="default")
+    except pittgoogle.exceptions.BadRequest as exc:
+        return str(exc), HTTP_400
+
+    blob = bucket.blob(_name_in_bucket(alert))
+    blob.metadata = _create_file_metadata(alert, event_id=envelope["message"]["messageId"])
+
+    # raise a PreconditionFailed exception if filename already exists in the bucket using "if_generation_match=0"
+    # let it raise. the message will be dropped.
```
troyraen:
```suggestion
```

This comment is a holdover from my original code but no longer makes sense because we've now implemented the try/except right here.

broker/cloud_run/swift/ps_to_storage/main.py
```diff
@@ -0,0 +1,107 @@
...
+
+def _name_in_bucket(alert: pittgoogle.Alert) -> str:
+    """Return the name of the file in the bucket."""
+    # not easily able to extract schema version, see:
+    # https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/swift/bat/Guano.example.json
+    _date = alert.dict["alert_datetime"][0:10]
+    _alert_type = alert.dict["alert_type"]
+    _id = alert.dict["id"][0]
+
+    return f"{_date}/{_alert_type}/{_id}.json"
```
troyraen: Let's figure out how to add the schema version. Otherwise it will be difficult to do things like figure out how the bucket organization maps to the dataset/table organization. The pipeline must know the version for things like naming the bigquery table, so perhaps the easiest solution is to add it as an env var to this module.

broker/cloud_run/swift/ps_to_storage/requirements.txt
```diff
@@ -0,0 +1,15 @@
...
+google-cloud-logging
+google-cloud-storage
+pittgoogle-client>=0.3.15
+
+# for Cloud Run
+# https://cloud.google.com/run/docs/quickstarts/build-and-deploy/deploy-python-service
+# pinned following quickstart example. [TODO] consider un-pinning
+Flask==3.0.3
+gunicorn==23.0.0
+Werkzeug==3.0.6
```
troyraen:
```suggestion
# for Cloud Run
# https://cloud.google.com/run/docs/quickstarts/build-and-deploy/deploy-python-service
Flask
gunicorn
Werkzeug
```

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -37,30 +38,27 @@ fi
 
 define_GCP_resources() {
     local base_name="$1"
+    local separator="$2"
```
troyraen:
```suggestion
    local separator="${2:--}"
```

And then update the calls below.

broker/setup_broker/swift/templates/bq_swift_policy.json
troyraen, line 1: This should be incorporated into our existing custom IAM setup instead of being a separate file. The roles and policies should be defined once when setting up the GCP project and then attached to individual resources when they are created. We manually created the roles (not sure about policies) on GCP a long time ago. https://github.com/mwvgroup/Pitt-Google-Broker/pull/147 was a later effort to update the role permissions, add the required files and code to the repo so that setup can be done programmatically, and also add docs explaining how to manage all the IAM stuff. Looks like it's basically done but I held it back wanting to finalize the student developer role. Familiarize yourself with that PR and then let's go over it during our next meeting. I hope we can finish the last To Do and merge during the meeting.

broker/setup_broker/swift/create_vm.sh
```diff
@@ -25,19 +27,22 @@ if [ "$teardown" = "True" ]; then
...
-    googlelogging="google-logging-enabled=true"
-    startupscript="startup-script-url=gs://${broker_bucket}/${survey}/vm_install.sh"
-    shutdownscript="shutdown-script-url=gs://${broker_bucket}/${survey}/vm_shutdown.sh"
-    gcloud compute instances create "$consumerVM" \
-        --zone="$zone" \
-        --machine-type="$machinetype" \
-        --scopes=cloud-platform \
-        --metadata="${googlelogging},${startupscript},${shutdownscript}"
+    if ! gcloud compute instances describe "${consumerVM}" --zone="${zone}" --project="${project_id}" >/dev/null 2>&1; then
+        machinetype=e2-custom-1-5632
```
troyraen, line 32: Do the installs actually succeed with a machine type this small? I recall needing a bigger machine for the set up and then making the machine type smaller for normal operations.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -37,30 +38,27 @@ fi
...
 
 #--- GCP resources used directly in this script
-broker_bucket=$(define_GCP_resources "${PROJECT_ID}-${survey}-broker_files")
-bq_dataset_alerts=$(define_GCP_resources "${survey}_alerts")
+artifact_registry_repo=$(define_GCP_resources "${survey}-cloud-run-services" "-")
+bq_dataset=$(define_GCP_resources "${survey}" "_")
+bq_table_alerts="alerts_${versiontag}"
+gcs_broker_bucket=$(define_GCP_resources "${PROJECT_ID}-${survey}-broker_files" "-")
+ps_subscription_alerts_reservoir=$(define_GCP_resources "${survey}-alerts-json-reservoir" "-")
+ps_topic_alerts_json=$(define_GCP_resources "${survey}-alerts-json" "-")
```
troyraen: Same as above, I don't think we want "-json" here.

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -72,55 +70,95 @@ manage_resources() {
...
-            --dead-letter-topic="${deadletter_topic_bigquery_import}" \
+            --dead-letter-topic="${ps_deadletter_topic}" \
             --max-delivery-attempts=5 \
             --dead-letter-topic-project="${PROJECT_ID}"
+        # set IAM policies on public Pub/Sub resources
+        if [ "$testid" = "False" ]; then
+            user="allUsers"
+            roleid="projects/${GOOGLE_CLOUD_PROJECT}/roles/userPublic"
+            gcloud pubsub topics add-iam-policy-binding "${ps_topic_alerts_json}" --member="${user}" --role="${roleid}"
+        fi
```
troyraen, line 139: 'alerts_raw' needs this same policy.

broker/cloud_run/swift/ps_to_storage/deploy.sh
```diff
@@ -0,0 +1,95 @@
...
+    echo
+    echo "Creating avro_bucket..."
+    if ! gsutil ls -b "gs://${gcs_avro_bucket}" >/dev/null 2>&1; then
+        #--- Create the bucket that will store the alerts
+        gsutil mb -l "${region}" "gs://${gcs_avro_bucket}"
+        gsutil uniformbucketlevelaccess set on "gs://${gcs_avro_bucket}"
+        gsutil requesterpays set on "gs://${gcs_avro_bucket}"
+        gcloud storage buckets add-iam-policy-binding "gs://${gcs_avro_bucket}" \
+            --member="allUsers" \
+            --role="roles/storage.objectViewer"
```
troyraen: Similar to the bigquery policy, we should incorporate the permissions being granted here into our custom userPublic role so that we consolidate all the permissions we intend to grant to the public and we can apply the same role/policy to every resource. Also, we should really only give public access to production resources.

broker/setup_broker/swift/create_vm.sh
```diff
@@ -2,15 +2,17 @@
...
-#   "False" uses production resources
-#   any other string will be appended to the names of all resources
-teardown="${3:-False}" # "True" tearsdown/deletes resources, else setup
-survey="${4:-swift}"
+# "True" tearsdown/deletes resources, else setup
+teardown="${3:-False}"
 # name of the survey this broker instance will ingest
+survey="${4:-swift}"
 zone="${5:-us-central1-a}"
+project_id="${6:-PROJECT_ID}"
```
troyraen: No need to define yet another name for this variable. Just use `$PROJECT_ID` directly where applicable. (GCP makes this variable confusing by using different names for it in different contexts -- PROJECT_ID and GOOGLE_CLOUD_PROJECT being most common, but there's at least one more. In our scripts we try to standardize on PROJECT_ID.)

broker/setup_broker/swift/setup_broker.sh
```diff
@@ -72,55 +70,95 @@ manage_resources() {
...
-
-        (cd templates && bq mk --table "${PROJECT_ID}:${bq_dataset_alerts}.${alerts_table}" "bq_${survey}_${alerts_table}_schema.json") || exit 5
-        bq update --description "Alert data from Swift/BAT-GUANO. This table is an archive of the swift-alerts Pub/Sub stream. It has the same schema as the original alert bytes, including nested and repeated fields." "${PROJECT_ID}:${bq_dataset_alerts}.${alerts_table}"
-
-        # create broker bucket and upload files
+        if ! bq ls "${PROJECT_ID}:${bq_dataset}" >/dev/null 2>&1; then
+            bq --location="${region}" mk --dataset "${bq_dataset}"
+            # grant public access to the dataset; for more information, see:
+            # https://cloud.google.com/bigquery/docs/control-access-to-resources-iam#grant_access_to_a_dataset
+            (cd templates && bq update --source "bq_${survey}_policy.json" "${PROJECT_ID}:${bq_dataset}") || exit 5
```
troyraen, line 81: Move this to the same block that grants public access to the pubsub topics so that it's only done for production resources.

broker/cloud_run/swift/ps_to_storage/deploy.sh
```diff
@@ -0,0 +1,95 @@
...
+        echo "Deleting resources for ${MODULE_NAME} module..."
+        gsutil rm -r "gs://${gcs_avro_bucket}"
+        gcloud pubsub topics delete "${ps_topic_avro}"
+        gcloud pubsub subscriptions delete "${ps_subscription_avro}"
+        gcloud pubsub subscriptions delete "${ps_input_subscrip}"
+        gcloud run services delete "${cr_module_name}" --region "${region}"
+    fi
+else
+    echo
+    echo "Creating avro_bucket..."
```
troyraen: Let's revisit which scripts create and delete which resources. I know we had a whole discussion about it when we first created these deploy.sh scripts for individual modules and decided to put this bucket creation here, but this is the only public resource that we handle this way and I've been confused by it more than once. Every other resource gets created by setup_broker.sh. At a minimum, our handling of this bucket and the bigquery dataset/table should be consistent.

troyraen (changes requested):
A few things to look at. Most important is (1) the "avro" resources; and (2) name of alert in bucket. I'm also noticing that the ways we assign permissions to users/resources has gotten pretty scattered. It would be good to consolidate and handle them consistently.

broker/cloud_run/swift/ps_to_storage/main.py
hernandezc1: @troyraen the alerts from Swift are JSON serialized. I named the Pub/Sub resource that way because at the time it seemed more appropriate and descriptive, but I agree that having a `<survey>-alerts` topic that is just a pass through of the survey's full (deduplicated) alert stream is the convention this module should adopt

broker/setup_broker/swift/create_vm.sh
hernandezc1: > Do the installs actually succeed with a machine type this small?

It worked when I tested it!

broker/cloud_run/swift/ps_to_storage/deploy.sh
hernandezc1: @troyraen the only solution I can think of is the following:
```
# add to line 43
ps_deadletter_topic_input_subscrip=$(define_GCP_resources "${survey}-upsilon-deadletter") 

# add to line 51
gcloud pubsub topics delete "${ps_deadletter_topic_input_subscrip}"

# add to line 61
gcloud pubsub topics create "${ps_deadletter_topic_input_subscrip}"

gcloud pubsub subscriptions create "${ps_input_subscrip}" \
  --topic="${ps_trigger_topic}" \
  --topic-project="${PROJECT_ID}" \
  --ack-deadline=600 \
  --push-endpoint="${url}${ROUTE_RUN}" \
  --push-auth-service-account="${runinvoker_svcact}" \
  --dead-letter-topic="${ps_deadletter_topic_input_subscrip}" \
  --max-delivery-attempts=5
```

Rather than retrying indefinitely, the message will be published to a dead letter topic after 5 delivery attempts

troyraen: Ok. What do you think about having only one deadletter topic that we use for every module? I'm resistant to having to manage a different one for each module but I also don't want to create a mess there if/when we actually need to use it. I don't have a clear sense of if/how we will actually want to work with messages that get delivered there. Do you? If we only have one deadletter topic and multiple modules start failing and dumping messages there, will we actually have a need to dig through them and sort out which ones came from which module? I think I would be more inclined to look at the bigquery tables to figure out which messages did/didn't make it through a given module. What do you think?

broker/cloud_run/swift/ps_to_storage/deploy.sh
hernandezc1: I'd like to standardize permissions across broker instances for all surveys in a separate PR

PR #279, opened 2025-06-27
Update deployment scripts for LSST

hernandezc1:
This PR standardizes the deployment scripts used to create or delete broker resources for LSST.

Primary goals:

- rename variables to improve the readability of the script
- improve documentation (e.g., the deployment scripts for Cloud Run services read `Cloud Functions` instead of `Cloud Run`)


broker/cloud_run/lsst/classify_snn/deploy.sh
```diff
@@ -10,109 +10,84 @@ teardown="${2:-False}"
...
 PROJECT_ID=$GOOGLE_CLOUD_PROJECT
-PROJECT_NUMBER=$(gcloud projects describe "$PROJECT_ID" --format="value(projectNumber)")
 
 MODULE_NAME="supernnova"  # lower case required by cloud run
 ROUTE_RUN="/"  # url route that will trigger main.run()
 
-# function used to define GCP resources; appends testid if needed
 define_GCP_resources() {
     local base_name="$1"
+    local separator="$2"
```
troyraen, line 21: Since we prefer "-" and only use "_" when required by GCP naming rules, you could set "-" as the default here and then most of the calls below wouldn't have to send this arg. That would make things less error prone, especially for future developers less familiar with our naming preferences..

troyraen (approved):
LGTM

PR #281, opened 2025-06-30
Add variable star classifier (UPSILoN) to ZTF Science Pipeline

hernandezc1:
This PR introduces a new module to Pitt-Google's ZTF science pipeline that uses [UPSILoN](https://github.com/dwkim78/upsilon) to identify variable sources and classify them.

Summary of changes:

- Outgoing messages from the ztf-variability module triggers a Cloud Run service (ztf-upsilon) to classify periodic variable stars
- Outgoing alerts are published to a Pub/Sub topic: ztf-upsilon
- Messages published to this topic are organized using two top level keys: alert_lite and upsilon.
- The alert_lite value contains the original fields of the lite alert packet; the upsilon value contains three outputs generated by UPSILoN for each band used to observe the object: label, probability, and flag as well as the objectId and candid associated with the alert packet.
- New Pub/Sub message attributes are included when these messages are published to the ztf-upsilon topic:
     - pg_upsilon_g_label, pg_upsilon_g_flag
     - pg_upsilon_r_label, pg_upsilon_r_flag
     - pg_upsilon_i_label, pg_upsilon_i_flag

A BigQuery subscription attached to the `ztf-upsilon` topic writes the upsilon data to a BigQuery table: upsilon.


broker/cloud_run/ztf/classify_upsilon/deploy.sh
```diff
@@ -0,0 +1,97 @@
...
+region="${4:-us-central1}"
+# get the environment variable
+PROJECT_ID=$GOOGLE_CLOUD_PROJECT
+
+MODULE_NAME="upsilon"  # lower case required by cloud run
+ROUTE_RUN="/"  # url route that will trigger main.run()
+
+define_GCP_resources() {
+    local base_name="$1"
+    local separator="${2:-}"
```
troyraen:
```suggestion
    local separator="${2:--}"
```

broker/cloud_run/ztf/classify_upsilon/deploy.sh
```diff
@@ -0,0 +1,97 @@
...
+    echo
+    echo "Creating trigger subscription for ${MODULE_NAME} Cloud Run service..."
+    # WARNING:  This is set to retry failed deliveries. If there is a bug in main.py this will
+    # retry indefinitely, until the message is delete manually.
+    gcloud pubsub subscriptions create "${ps_input_subscrip}" \
+        --topic "${ps_trigger_topic}" \
+        --topic-project "${PROJECT_ID}" \
+        --ack-deadline=600 \
+        --push-endpoint="${url}${ROUTE_RUN}" \
+        --push-auth-service-account="${runinvoker_svcact}"
```
troyraen: Did you figure out how to limit the number of retries for cloud run services? If so, implement it here so we can drop this disaster waiting to happen 🥴.

broker/cloud_run/ztf/classify_upsilon/main.py
```diff
@@ -0,0 +1,139 @@
...
+        TOPIC.publish(
+            pittgoogle.Alert.from_dict(
+                {
+                    "alert_lite": alert_lite.dict,
+                    "upsilon": {
+                        "objectId": alert_lite.dict["alertIds"]["objectId"],
+                        "candid": alert_lite.dict["alertIds"]["sourceId"],
+                        **upsilon_dict,
+                    },
+                },
```
troyraen: I don't think we need to duplicate the object and source IDs in the value added dicts.

Does this also need to pass on the variability dict?

broker/cloud_run/ztf/classify_upsilon/main.py
```diff
@@ -0,0 +1,139 @@
...
+    has_min_detections_in_any_band = any(
+        alert_lite.dict["variability"].get(f"n_detections_{band}_band", 0) >= 80
+        for band in ["g", "r", "i"]
+    )
+
+    if has_min_detections_in_any_band:
+        # classify and publish results
+        alert_lite_df = _create_lite_dataframe(alert_lite.dict["alert_lite"])
+        upsilon_dict = _classify_with_upsilon(alert_lite_df)
+        TOPIC.publish(
```
troyraen: Same as for LSST -- repeating that comment here... Putting these classify and publish statements underneath `if` means that this module filters the alert stream. I don't think we want that. Our philosophy so far has been to run every alert through every module. That does lead to silly things like producing variable star classifications for known asteroids and supernovae, but the benefit is that we don't drop edge cases before users have had a chance to look at them. Completeness over purity. I recommend running upsilon on every alert but adding a flag to `upsilon_dict` and the outgoing message attributes that indicates whether the alert input met the minimum recommended criteria. Since min number of detections is the only criteria here, maybe call it 'has_min_detections' with values 0=False, 1=True.

Will also note that if we really do want a module to process only some alerts then we should implement that filter at the subscription level when possible rather than paying for those alerts to trigger the module only to be immediately dropped.

broker/cloud_run/ztf/classify_upsilon/main.py
```diff
@@ -0,0 +1,139 @@
...
+        upsilon_dict[f"{band}_flag"] = flag
+
+    return upsilon_dict
+
+
+def _create_lite_dataframe(alert_dict: dict) -> pd.DataFrame:
+    """Return a pandas DataFrame containing the source detections."""
+
+    # sources and previous sources are expected to have the same fields
+    sources_df = pd.DataFrame([alert_dict.get("source")] + (alert_dict.get("prvSources") or []))
```
troyraen:
```suggestion
    sources_df = pd.DataFrame([alert_dict.get("source")] + (alert_dict.get("prv_candidates") or []))
```

Double check this using an alert that has both of these.

broker/cloud_run/ztf/classify_upsilon/main.py
```diff
@@ -0,0 +1,139 @@
...
+    # use nullable integer data type to avoid converting ints to floats for columns in one dataframe but not the other
+    sources_ints = [c for c, v in sources_df.dtypes.items() if v == int]
+    dataframe = sources_df.astype({c: "Int64" for c in sources_ints})
```
troyraen: Looking at the code comments, this should not be necessary because sources and previous sources are expected to have the same fields. If it is necessary, it must be done before creating the dataframe. Otherwise, the ints will already have been converted to floats. Converting after the fact will not reliably recover the original values.

broker/setup_broker/ztf/templates/bq_ztf_upsilon_schema.json
```diff
@@ -0,0 +1,76 @@
+[
+    {
+      "description": "Classification results using UPSILoN (Kim & Bailer-Jones 2015).",
+      "fields": [
+          {
+            "description": "object identifier or name.",
```
troyraen:
```suggestion
            "description": "Object identifier or name.",
```

broker/cloud_run/ztf/classify_upsilon/main.py
```diff
@@ -0,0 +1,139 @@
...
+            )
+        )
+
+    return "", HTTP_204
+
+
+def _classify_with_upsilon(alert_lite_df: pd.Dataframe) -> dict:
+    upsilon_dict = {}
+    bands = alert_lite_df["filter"].map(pittgoogle.utils.ztf_fid_names()).unique()
+    rf_model = upsilon.load_rf_model()  # load UPSILoN's classification model
```
troyraen: Would it be better to put this line at module level instead of function level?

troyraen (approved):
Check a few things below. Most importantly, the location of the classify and publish statements.

PR #282, opened 2025-07-01
Update `lsst-upsilon` module

hernandezc1:
This PR updates the `lsst-upsilon` Cloud Run service.

**Summary of changes**
- the Pub/Sub topic that triggers this module is now `lsst-variability` (previously `lsst-lite`)
     - "For the best classification quality, [UPSILoN] recommend[s] to use light curves with more than ~80 data points." The "value_added" information provided by the `lsst-variability` module contains the number of data points in every LSST band used to observe the diaSources associated with a diaObject in each alert packet.
- Before a classification is made, the Python script now checks to see if any of the bands contain at least 80 data points


broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -22,23 +22,21 @@
 TESTID = os.getenv("TESTID")
 SURVEY = os.getenv("SURVEY")
 
-# Variables for incoming data
+# ---Variables for incoming data
 # A url route is used in setup.sh when the trigger subscription is created.
 # It is possible to define multiple routes in a single module and trigger them using different subscriptions.
 ROUTE_RUN = "/"  # HTTP route that will trigger run(). Must match deploy.sh
+survey_bands = ["u", "g", "r", "i", "z", "y"]
```
troyraen: Upper-case this variable name to indicate a module-level constant.

troyraen (approved):
LGTM

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -62,78 +60,82 @@ def run() -> tuple[str, int]:
...
+        upsilon_dict = _classify_with_upsilon(alert_lite_df)
+        TOPIC.publish(
+            pittgoogle.Alert.from_dict(
+                {
+                    "alert_lite": alert_lite.dict,
+                    "upsilon": {
+                        "diaObjectId": alert_lite.dict["diaObject"]["diaObjectId"],
+                        "diaSourceId": alert_lite.dict["diaSource"]["diaSourceId"],
+                        **upsilon_dict,
+                    },
```
troyraen: Just realized, if this module is triggered by the 'variability' stream, should there be another key in here passing that dict on?

broker/cloud_run/lsst/classify_upsilon/deploy.sh
```diff
@@ -18,7 +18,7 @@ ROUTE_RUN="/"  # url route that will trigger main.run()
 
 define_GCP_resources() {
     local base_name="$1"
-    local separator="$2"
+    local separator="${2:-}"
```
troyraen:
```suggestion
    local separator="${2:--}"
```

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -62,78 +60,82 @@ def run() -> tuple[str, int]:
...
+    if has_min_detections_in_any_band:
+        alert_lite_df = _create_lite_dataframe(alert_lite.dict["alert_lite"])
+        upsilon_dict = _classify_with_upsilon(alert_lite_df)
+        TOPIC.publish(
+            pittgoogle.Alert.from_dict(
+                {
+                    "alert_lite": alert_lite.dict,
+                    "upsilon": {
+                        "diaObjectId": alert_lite.dict["diaObject"]["diaObjectId"],
+                        "diaSourceId": alert_lite.dict["diaSource"]["diaSourceId"],
```
troyraen: I don't think we need to duplicate these in the value added dicts.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -62,78 +60,82 @@ def run() -> tuple[str, int]:
...
+    has_min_detections_in_any_band = any(
+        alert_lite.dict["variability"].get(f"n_detections_{band}_band", 0) >= 80
+        for band in survey_bands
     )
 
-    return "", HTTP_204
+    if has_min_detections_in_any_band:
+        alert_lite_df = _create_lite_dataframe(alert_lite.dict["alert_lite"])
+        upsilon_dict = _classify_with_upsilon(alert_lite_df)
+        TOPIC.publish(
```
troyraen: Sorry, missed this the first time around as well... Putting these classify and publish statements underneath `if` means that this module filters the alert stream. I don't think we want that. Our philosophy so far has been to run every alert through every module. That does lead to silly things like producing variable star classifications for known asteroids and supernovae, but the benefit is that we don't drop edge cases before users have had a chance to look at them. Completeness over purity. I recommend running upsilon on every alert but adding a flag to `upsilon_dict` and the outgoing message attributes that indicates whether the alert input met the minimum recommended criteria. Since min number of detections is the only criteria here, maybe call it 'has_min_detections' with values 0=False, 1=True.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -62,78 +60,82 @@ def run() -> tuple[str, int]:
...
+        upsilon_dict[f"{band}_label"] = label
+        upsilon_dict[f"{band}_probability"] = probability
+        upsilon_dict[f"{band}_flag"] = flag
 
-    return outgoing_dict
+    return upsilon_dict
 
 
-def _create_dataframe(alert_dict: pittgoogle.Alert) -> pd.DataFrame:
+def _create_lite_dataframe(alert_dict: pittgoogle.Alert) -> pd.DataFrame:
```
troyraen:
```suggestion
def _create_lite_dataframe(alert_dict: dict) -> pd.DataFrame:
```

Right? If not, the keys being used below in `alert_dict.get()` are wrong.

broker/cloud_run/lsst/classify_upsilon/main.py
```diff
@@ -62,78 +60,82 @@ def run() -> tuple[str, int]:
...
-        "diaObjectId": alert_lite_dict["diaObject"]["diaObjectId"],
-        "diaSourceId": alert_lite_dict["diaSource"]["diaSourceId"],
-    }
 
-    for band in bands:
-        # define parameters for feature extractions
-        filter_diaSources = alert_df[alert_df["band"] == band]
+def _classify_with_upsilon(alert_lite_df: pd.DataFrame) -> dict:
+    upsilon_dict = {}
+    rf_model = upsilon.load_rf_model()  # load UPSILoN's classification model
```
troyraen: I see this was moved from module level to function level. Was that on purpose? Module level seems more appropriate but it might not make a significant difference either way.

PR #283, opened 2025-07-01
Fix bug in LSST's setup_broker.sh script

hernandezc1:
This PR fixes a bug in the setup_broker.sh script that was incorrectly defining GCP resource names. The deploy.sh script for each of Pitt-Google's LSST Cloud Run services was also updated to include this new change.


troyraen (approved):
LGTM

PR #299, opened 2025-09-16
Update LSST Cloud Run services

hernandezc1:
### Changed

- updated `main.py` for the `classify_snn` and `classify_upsilon` modules
     - These scripts now return `None` values in the value added dictionaries (e.g., `snn_dict`, `upsilon_dict`) for alerts with sources identified as SS objects
- updated `main.py` for the `ps_to_storage` module
     - Included `kafka.timestamp` to the alert metadata
- updated `deploy.sh` for every Cloud Run service
     - the `pubsub.subscriber` role is now added to the push subscription trigger
- `requirements.txt` for every Cloud Run service has been updated to use v0.3.18 or greater of the pittgoogle-client


troyraen (approved):
LGTM

PR #302, opened 2025-10-07
Update SMT UDFs for LSST modules

hernandezc1:
### Changed

- Cloud Run services use `pittgoogle-client>=0.3.19`
- Defined functions in `broker/cloud_run/lsst/ps_to_storage/main.py` to generate lite alerts before they are published to `lsst-lite`
- `broker/setup_broker/lsst/setup_broker.sh`
     - remove `ps_lsst_lite_smt.yaml` from `ps_topic_alerts_lite`
- Updated the UDF in the following SMT files to prevent loss of data precision
     - `ps_lsst_flatten_schema_smt.yaml` and `ps_lsst_add_top_level_fields_smt.yaml` 
- Updated the description for the field `diaSourceId` in value added BigQuery table schemas
     - This field matches the description for the field `diaSource.diaSourceId` in the [alert packet schema](https://github.com/lsst/alert_packet/blob/b1ecb9c006d211d8117f4672d4044ff8fca22d7e/python/lsst/alert/packet/schema/9/0/lsst.v9_0.diaSource.avsc#L7)
- `ps_input_subscription` for our value added modules are now configured with [exponential backoffs](https://cloud.google.com/pubsub/docs/subscription-retry-policy#exponential_backoff) for delivery failures
     - "While using exponential backoff, Pub/Sub continues to deliver other messages, even if previous messages received negative acknowledgments (unless, you're using ordered message delivery)."

### Removed

- `broker/setup_broker/lsst/templates/ps_lsst_lite_smt.yaml`


broker/cloud_run/lsst/classify_snn/deploy.sh
```diff
@@ -92,7 +92,9 @@ else
         --push-endpoint="${url}${ROUTE_RUN}" \
         --push-auth-service-account="${runinvoker_svcact}" \
         --dead-letter-topic="${ps_deadletter_topic}" \
-        --max-delivery-attempts=5
+        --max-delivery-attempts=5 \
+        --min-retry-delay=10 \
+        --max-retry-delay=600
```
troyraen, line 97: At some point, double check that the combination of these three flags does what you expect.

broker/cloud_run/lsst/ps_to_storage/main.py
```diff
@@ -109,3 +151,33 @@ def _create_file_metadata(alert: pittgoogle.Alert, event_id: str) -> dict:
...
+            # replace the original nested object with its filtered version
+            alert_lite_dict[key] = _process_field(alert_lite_dict.get(key), config)
+
+    return pittgoogle.Alert.from_dict(
+        payload={"alert_lite": alert_lite_dict},
+        attributes={**alert.attributes},
+    )
+
+
+def _process_field(original_value: Any, config: dict) -> Any:
```
troyraen: Can this really accept and return `Any` type? Looks to me like it only handles `dict | (list[dict] | None)`.

broker/cloud_run/lsst/ps_to_storage/main.py
```diff
@@ -28,6 +29,51 @@
 # Variables for outgoing data
 HTTP_204 = 204  # HTTP code: Success
 HTTP_400 = 400  # HTTP code: Bad Request
+LITE_FIELDS_CONFIG = {
+    "diaSource": {
+        "fields": {
+            "diaSourceId",
```
troyraen, line 35: Fine for now, but it would be great to define this in a yaml file in pittgoogle-client instead. `_create_lite_alert()` could move there too.

broker/setup_broker/lsst/templates/ps_smt_add_top_level_fields.yaml
```diff
@@ -1,21 +1,39 @@
 # https://cloud.google.com/pubsub/docs/smts/create-topic-smt#create
 - javascriptUdf:
     code: >
-        function addTopLevelFields(message, metadata) {
-          const data = JSON.parse(message.data);
-          const attrs = message.attributes || {};
+      function addTopLevelFields(message, metadata) {
+        const attrs = message.attributes || {};
+        const dataStr = message.data.toString();
```
troyraen, line 6: Does `toString()` have any effect on the data that comes out?

broker/setup_broker/lsst/templates/ps_smt_add_top_level_fields.yaml
```diff
@@ -1,21 +1,39 @@
...
 
-          return {
-            data: JSON.stringify(payload),
-            attributes: attrs // preserve attributes
-          };
+        // Extract the following attributes and add them to newFields
+        // We avoid casting fields as JavaScript numbers to prevent precision loss
+        if (attrs.healpix9)  newFields.healpix9  = attrs.healpix9.toString();
+        if (attrs.healpix19) newFields.healpix19 = attrs.healpix19.toString();
+        if (attrs.healpix29) newFields.healpix29 = attrs.healpix29.toString();
```
troyraen, line 15: Does javascript really not support 64bit ints? That seems strange. If leaving `toString()`, does that mean these will be strings in the output data? That would be a pain for everyone downstream.

broker/setup_broker/lsst/templates/ps_smt_flatten_schema.yaml
```diff
@@ -1,49 +1,29 @@
...
+        const valueAddedData = valueAddedKey ? (data[valueAddedKey] || {}) : {};
+
+        // Build the final payload, starting with value_added fields
+        const payload = {
+          ...valueAddedData,
+
+          // Use the attributes to add the remaining key-value pairs
+          diaSourceId: attrs.diaSource_diaSourceId ? attrs.diaSource_diaSourceId.toString() : null,
+          diaObjectId: attrs.diaObject_diaObjectId ? attrs.diaObject_diaObjectId.toString() : null,
+          ssObjectId: attrs.ssSource_ssObjectId ? attrs.ssSource_ssObjectId.toString() : null,
```
troyraen, line 19: Same concern about `toString()` here.

broker/setup_broker/lsst/templates/ps_smt_add_top_level_fields.yaml
troyraen, line 1: Add comments at the tops of these SMT yaml files briefly explaining what the transforms do and where there are intended to be used.

Also, minor, but I would name these files like `ps_smt_lsst_add_top_level_fields.yaml` or just `smt_lsst_add_top_level_fields.yaml` so "smt" is closer to the beginning. Also, 'lsst' could probably be removed since these are underneath an 'lsst' directory.

broker/setup_broker/lsst/templates/ps_smt_add_top_level_fields.yaml
```diff
@@ -1,21 +1,39 @@
...
+        if (attrs["kafka.timestamp"]) {
+          newFields.kafkaPublishTimestamp = attrs["kafka.timestamp"] * 1000;
```
troyraen, line 17: 'kafka.timestamp * 1000' is happening in several places (2 in this PR and I think I saw it somewhere else when you were screensharing earlier today). Is it necessary? If so, leave a comment explaining why. 

Also consider whether a user could be easily confused when comparing the data that comes out of a '* 1000' operation with data from the same alert that hasn't undergone '* 1000'. In most cases I'm aware of where a '* 1000' operation is necessary, the transformation results in output data will not be confusing (e.g., the output is a python `datetime` object, so not a float like 'kafka.timestamp' is and thus not confusing). I don't know what the output data looks like here, which is why I suggest considering it. If it is both necessary and confusing, what to do about that is probably a case-by-case decision.

troyraen (approved):
Thanks!

PR #303, opened 2025-10-07
Optimize Python scripts for LSST Cloud Run services

hernandezc1:
### Updates

- `broker/cloud_run/lsst/.../main.py`
     - the `_create_dataframe` function in `main.py` is now optimized to only include the fields required as input for the classifier (or module in the case of `variability`). This will reduce the amount of memory needed to process a single alert


troyraen (approved):
LGTM

